PR #136, opened 2026-03-17
feat(analyzer): add 5 error/perf rules and enhance DEFERCLOSE (360 total)

kodflow:
## Summary

Add 5 new lint rules and enhance 1 existing rule, bringing the total from 355 to 360.

### New rules
- **KTN-ERROR-DEFERCLOSE** (enhanced): detects closure-wrapped error ignore patterns (`defer func() { _ = x.Close() }()`)
- **KTN-ERROR-CLOSECHECK**: detects `Close()/Shutdown()/Stop()` outside defer with ignored error
- **KTN-ERROR-DISCARD**: detects any error assigned to `_` — must log or propagate
- **KTN-FUNC-NILRETURN**: detects functions where a return value is always the same constant (nil/true/false)
- **KTN-VAR-DEFERPERF**: detects unnecessary `defer Close` when immediate close is more performant

### Bug fixes
- Fixed `NumFields()` grouped return bug in NILRETURN (`(a, b error)` → 1 field but 2 values)
- Replaced all `_ = cerr` patterns with `log.Printf` (errors must never be silently swallowed)
- Fixed Cobra flag error handling in `config.go`

### Test improvements
- Added `t.Parallel()` to all applicable test functions
- Replaced silent `return` with `t.Fatal()` in test nil checks
- Added `messages.Get()` fallback when message not registered

## Changes

| Category | Files | Details |
|----------|-------|---------|
| New rules | 5 | closecheck.go, discard.go, deferperf.go, deferclose enhancement, nilreturn.go |
| Tests | 12 new | External + internal tests + testdata (good.go/bad.go) per rule |
| Self-lint fixes | 8 | cache, updater, cmd — `_ = cerr` → `log.Printf` |
| Registry | 5 | messages, severity, handling/registry, variable/registry |
| Docs | 4 | README.md, CLAUDE.md files — 153 KTN + 27 + 20 + 160 = 360 total |

## Test plan

- [x] `make build` compiles
- [x] `make test` — all packages pass, 0 failures
- [x] Each rule's `bad.go` produces only expected diagnostics
- [x] Each rule's `good.go` produces 0 diagnostics
- [x] Self-lint: 0 violations for new rules on project code
- [x] CodeRabbit review: all actionable findings addressed


kodflow (commented):
Addressed all actionable Qodo findings. The 2 remaining threads are intentional design decisions (testdata uses log.Println by design, and types.String() comparison is pragmatic for stable stdlib types).

PR #170, opened 2026-04-02
fix(serve): resilient MCP server with infinite retry and proxy mode

kodflow:
## Summary

Make all MCP server components self-healing so the server never crashes permanently:

- **HTTP**: bind failure retries with exponential backoff (1s→30s) until port available
- **Watcher**: fsnotify crash detected via `Stopped()` channel, supervisor restarts automatically
- **Socket**: creation failure retries with backoff
- **Singleton**: proxy mode via Unix socket instead of `exit()` — prevents MCP handshake failure

## Problem

When any component failed at startup (port taken, OS resource limit, stale socket), the entire MCP server would exit. This caused the parent process to mark it as "failed" with no retry.

## Solution

| Component | Before | After |
|-----------|--------|-------|
| HTTP bind fails | **Server exits** | Retry with backoff, MCP continues |
| Watcher crash | **Silent death** | Supervisor detects + restarts |
| Socket creation fails | **Non-fatal warning** | Retry with backoff |
| Another instance running | **exit(0) — no handshake** | Proxy MCP via socket to existing |

## Changed files

- `pkg/serve/serve.go` — retry/supervisor infrastructure with `retryComponent` + `superviseWatcher`
- `pkg/serve/mcp.go` — `NewProxyMCPServer` for socket delegation
- `pkg/serve/watcher.go` — `Stopped()` channel for crash detection
- `pkg/serve/engine_manager.go` — `MainWatcher()` accessor
- `cmd/ktn-linter/cmd/serve.go` — `runProxy()` singleton handler

## Test plan

- [x] `go build ./...` passes
- [x] 1576 tests pass (`pkg/serve` + `cmd/ktn-linter/cmd`)
- [x] Zero new phase 1-6 lint violations
- [x] Pre-existing tests adapted to new `startComponents()` signature

- CLI behavior: singleton detection in runServe() now enters proxy mode (runProxy) instead of exiting; runProxy resolves --max-phase (falls back to config), installs SIGTERM/SIGINT + parent watchdog, and delegates to serve.RunProxy to maintain an stdio-based MCP handshake while proxying scan/status calls to the running instance.
- Server resilience / orchestrator changes: startup failures are non-fatal and self-healing — HTTP and Unix socket binds retry indefinitely with exponential backoff (1s → 30s), watcher crashes are detected via ServerWatcher.Stopped() and auto-restarted by a supervisor loop, and Server now holds an internal ctx to govern retry/supervisor lifetimes and orderly shutdown.
- Proxy MCP and API changes: added NewProxyMCPServer and RunProxy; proxy tools (scan, status) validate inputs, forward calls to the client, return errors as mcp.CallToolResult.IsError=true, and filter scan results by FilterByPhase(maxPhase). Also added EngineManager.MainWatcher() for main watcher lookup.
- Tests / modernization / registry: CI: go build ./... and 1576 tests pass; tests adapted for non-fatal startComponents (context injection, explicit cancel, s.socketSrv.Close(), wg sync). No changes to severity/message registry or new KTN rule codes introduced; one modernization/govet-type improvement: retryBackoffMultiplier changed to time.Duration.


kodflow (commented):
All CodeRabbit findings triaged and fixed in current commit (e8a5f1ed):

1. **Signal goroutine leak** → Fixed: `select` on both `sigCh` and `ctx.Done()` with `signal.Stop(sigCh)` cleanup
2. **Redundant loadConfiguration** → Fixed: removed from `runProxy`, config loaded once before singleton check
3. **retryBackoffMultiplier type** → Fixed: changed to `time.Duration`, removed explicit casts

Additionally fixed Qodo finding:
4. **Shutdown deadlock** → Fixed: `runCancel()` called before `drain()/shutdown()` via derived internal context

PR #180, opened 2026-04-09
fix(multi): resolve 12 consensus bugs from codebase audit

kodflow:
## Summary

- Fix modernize prefix mismatch (`KTN-MODERNIZE-` → `KTN-MDRNZ-`) across phases, generator, and tests
- Fix `CheckOperandUsesAnyType` to validate `anyTypeParams` mapping (was ignoring the parameter)
- Add nil `Fset` guard in `DiagnosticResult.Position()` to prevent panic on broken packages
- Cache wrapped analyzers via `sync.Once` in staticcheck (160) and govet (27) registries
- Align `Extract()` dedup key with `Filter()` (prefix before dedup, not after)
- Remove TOCTOU race in `RemoveEngine` (nil assignment before delete)
- Fix `Logger.write()` lock discipline (`RLock` → `Lock` for write operations)
- Remove dead `reparseStartMsg` type and handler
- Fix watcher close race via `defer Close()` after `Start()`
- Always watch config directory even if YAML file doesn't exist yet
- Add `childMap` to `TreeNodeValue` for O(1) `findChild` during tree construction

## Test plan

- [ ] `make test` passes (19,000+ tests)
- [ ] `make build` compiles cleanly
- [ ] Modernize rules correctly classified as `PhaseModern` (not `PhaseStyle`)
- [ ] No panic when analyzer produces diagnostic with nil Fset
- [ ] Tree rebuild performance improved on large projects

- Modernize rule codes renamed: `KTN-MODERNIZE-...` → `KTN-MDRNZ-...` across generator, phase mapping, and tests (e.g. tests now expect `KTN-MDRNZ-STRCUT`); modernize rules continue to classify as PhaseModern.

- Analyzer registry caching & severity handling: staticcheck (~160 analyzers) and govet (27 analyzers) now build and cache their wrapped analyzer slices with `sync.Once` (avoids repeated wrap/allocation). `registerSeverities()` is invoked in the single init path so severities are registered once.

- Orchestrator / diagnostics hardening and cache/orchestration behavior: `DiagnosticsProcessorService.Extract()` applies the external KTN prefix before dedup-key generation (aligning with `Filter()`), `DiagnosticResult.Position()` guards against nil `Fset` to prevent panics, and engine/watcher removal eliminated a TOCTOU by deleting map entries directly.

- CLI/config and logger fixes: watcher lifecycle now defers `Close()` after `Start()` to fix a close race and the server watcher no longer skips watching the config directory (so new `.ktn-linter.yaml` files are detected). `LoggerService.write()` changed from `RLock` to `Lock` to enforce exclusive writes.

- Tree construction & tests: `TreeNodeValue` gained a `childMap` for O(1) child lookup (improves rebuild performance on large projects). Tests updated (removed `reparseStartMsg` type and related tests); full test and build targets remain expected to pass (`make test` ≈19k tests, `make build`).


kodflow:
## Review findings triage — 7 fixed, 1 rejected

### Fixed (commit e45fb243)
- **CR: Redundant severityOnce** — removed, registerSeverities() called directly from initOnce
- **CR: Stale KTN-MODERNIZE comments** — updated to KTN-MDRNZ in generator.go
- **CR: Redundant configDir watch** — rootDir already watched by addDirectories
- **CR: CloseAll nil-then-delete** — consistent with RemoveEngine pattern
- **CR: runFromCachePolling defer watcher** — applied defer Close() pattern
- **CR: runCacheTUI defer watcher** — same pattern
- **CR: childMap sync** — insertDirPath already populates childMap, BuildTree only calls insertDirPath (verified)

### Rejected with justification
- **CR: Defensive copy of cachedAnalyzers** — callers never mutate the returned slice (read-only iteration in selector.go and runner.go). Defensive copy allocates 160+ pointers per call, negating the caching optimization.

@coderabbitai resolve

kodflow:
@coderabbitai full review

PR #181, opened 2026-04-09
fix(lint): resolve RWOVERMU false positives + lint violations

kodflow:
## Summary

- Fix RWOVERMU false positives: rule now detects `delete(recv.field, key)` and `recv.Method()` as write operations
- Add `hasDeleteOnReceiverFieldRwovermu` and `isReceiverMethodCallRwovermu` detection helpers with testdata
- Fix KTN-VAR-GROUP in `pkg/explore` by merging vars into single `var()` block
- Fix KTN-COMMENT-BLOCK/FUNC in `pkg/explore`, `pkg/serve` by adding `//: ` intention comments

## Test plan

- [ ] `go test ./pkg/analyzer/ktn/goroutine/lockfree/...` — 82 RWOVERMU tests pass
- [ ] `go test ./pkg/explore/... ./pkg/serve/...` — 3614 tests pass across modified packages
- [ ] `ktn-linter prompt ./...` — 0 non-test-quality violations remain
- [ ] RWOVERMU testdata `good.go` has 0 RWOVERMU violations (new delete + receiver method cases)


kodflow:
**Qodo P1 finding triaged — acknowledged, tracked as follow-up**

The inconsistency between `explore.isExternalCachePath` (13 segments) and `orchestrator.isExternalPath` (2 segments) is a valid observation. However, this is pre-existing behavior — this PR only moved `externalPathSegments` into a `var()` block and added `//: ` intention comments. The expanded segment list was already present before this PR.

The proper fix (shared `pathfilter.IsExternalPath` helper) is a separate concern and will be addressed in a follow-up. Not blocking this PR as it introduces no regression.

PR #187, opened 2026-04-21
fix(serve): apply FilterDiagnostics in scanAllMultiModule (#186)

### LINKED ISSUE #186: MCP `serve` cache bypasses `FilterDiagnostics` on multi-module workspaces → CLI and MCP scan disagree on `*_test.go` files
## Summary

`scanAllMultiModule` in the MCP `serve` engine skips `orchestrator.FilterDiagnostics`, so the cache that feeds the MCP `scan`/`dump` tools (and therefore the `dev` TUI) contains diagnostics that `ktn-linter lint ./...` silently discards via the built-in *test-file* exclusion list.

On a multi-module workspace (any repo with a `go.work`), the two entry points disagree:

| Entry point | Applies `FilterDiagnostics`? | Result on a multi-module repo |
|---|---|---|
| `ktn-linter lint ./...` (CLI) | ✅ yes | `No issues found! Code is compliant.` |
| `ktn-linter serve` → MCP `scan` / `dump` → `dev` TUI | ❌ no | 23 issues reported on `*_test.go` files |

All 23 reported issues fall under rules that are in `rulesWithBuiltinTestExclusion` (KTN-COMMENT-FUNC, KTN-VAR-NILSLICE, KTN-FUNC-NAMEDPARAMS, KTN-VAR-MAKEAPPEND, KTN-FUNC-NOMAGIC, …) — the exact set that the CLI correctly filters out by default.

## Version

- ktn-linter **v1.3.163** (installed as `/home/vscode/.cache/go/bin/ktn-linter`, `go install github.com/kodflow/ktn-linter/cmd/ktn-linter@latest` from Go 1.26.2)
- OS: Linux 6.12 (Debian-based devcontainer)
- Workspace: multi-module Go project with a top-level `go.work` (4 modules: `internal/kernel`, `internal/core`, `internal/service`, `pkg/v1`)

## Reproduction

A minimal reproducer needs: a `go.work` with ≥ 2 modules, and any `*_test.go` file that would trigger a rule in the built-in test-exclusion list (e.g. an unnamed return parameter → `KTN-FUNC-NAMEDPARAMS`, or a missing doc comment → `KTN-COMMENT-FUNC`).

Observed on a real workspace:

```
$ ktn-linter lint ./...
✅ No issues found! Code is compliant.

$ ktn-linter            # dev TUI fed by the MCP cache
KTN-Linter dev                                         23 issues [h]ide: ON ● live ● MCP (cached)
[ 99.7%] ● workspace/                          23 ● ● ●
[ 99.6%] ├── ● internal/                       22
[100.0%] │   ├── ● core/codec/
           ╰── registry_external_test.go       9
[100.0%] │   ├── ● kernel/buffer/
           ╰── buffer_external_test.go         2
[100.0%] │   ╰── ● service/
             ├── codec/ndjson/                 4
             ├── codec/yaml/                   1
             ╰── logger/                       6
[100.0%] ╰── ● pkg/v1/logger/
             ╰── logger_external_test.go       1
```

Every one of the 23 issues is inside a `*_test.go` file. The issues are distributed across the "Style & Naming" (19), "Performance" (1) and "Logic & Correctness" (3) categories — rule codes that all appear in `rulesWithBuiltinTestExclusion`.

## Expected vs actual

**Expected:** the MCP cache (and therefore the `scan` / `dump` tools and the `dev` TUI) reports the same set of violations as the CLI after default exclusions, because both rely on the same `orchestrator.FilterDiagnostics` pipeline.

**Actual:** on multi-module workspaces the MCP cache contains raw diagnostics, including those suppressed by `rulesWithBuiltinTestExclusion` / `testdataExcludedRules` / `isExternalPath`. Single-module workspaces are unaffected.

## Root cause

The divergence is between the CLI's multi-module pipeline and the engine's multi-module scan:

**CLI — `cmd/ktn-linter/cmd/lint.go` (`runMultiModulePipeline`):**
```go
rawDiags, pkgErrors, err := orch.RunMultiModule(args, opts)
if err != nil { ... }
filtered := orch.FilterDiagnostics(rawDiags)          // ✅ applies built-in test/testdata exclusions
return filtered, pkgErrors, nil
```

**MCP engine — `pkg/serve/engine.go:477`, `scanAllMultiModule`:**
```go
func (e *EngineService) scanAllMultiModule(patterns []string) (result *ScanResult) {
    orch := orchestrator.NewOrchestratorService(e.stderr, e.verbose)
    searchPaths := e.resolveSearchPaths(patterns)

    allDiags, _, runErr := orch.RunMultiModule(searchPaths, e.opts)
    if runErr != nil { ... }

    issues := diagnosticsToIssues(allDiags)            // ❌ no FilterDiagnostics call
    pkgPaths := collectPackageDirsFromDiags(allDiags)
    ...
}
```

`orchestrator.RunMultiModule` (`pkg/orchestrator/orchestrator.go:251`) returns raw diagnostics by design — it deliberately leaves filtering to the caller. The CLI does it, the engine does not.

Note that the **other** two scan paths in the same engine *do* filter correctly:

- `scanAllSingleModule` (`pkg/serve/engine.go:577`): `filtered := orch.FilterDiagnostics(rawDiags)`
- `analyzeAndCache` / per-package scan (`pkg/serve/engine.go:1140`): `filtered := orch.FilterDiagnostics(rawDiags)`

Only `scanAllMultiModule` was forgotten. Since `ScanAll` is the warm-up path (`populateCacheFromScanAll`, `pkg/serve/engine.go:631`), every subsequent `scan`/`dump` call serves the unfiltered cache.

## Proposed fix

Mirror what the single-module path already does:

```diff
--- a/pkg/serve/engine.go
+++ b/pkg/serve/engine.go
@@ -478,13 +478,14 @@ func (e *EngineService) scanAllMultiModule(patterns []string) (result *ScanResul
 	orch := orchestrator.NewOrchestratorService(e.stderr, e.verbose)
 	searchPaths := e.resolveSearchPaths(patterns)

 	allDiags, _, runErr := orch.RunMultiModule(searchPaths, e.opts)
 	//: Multi-module scan can fail if no modules are found or on filesystem errors.
 	if runErr != nil {
 		e.logVerbose("ScanAll multi-module error: %v\n", runErr)
 		//: Return the computed result to the caller.
 		return &ScanResult{ScannedAt: time.Now()}
 	}

-	issues := diagnosticsToIssues(allDiags)
-	pkgPaths := collectPackageDirsFromDiags(allDiags)
+	filtered := orch.FilterDiagnostics(allDiags)
+	issues := diagnosticsToIssues(filtered)
+	pkgPaths := collectPackageDirsFromDiags(filtered)
 	now := time.Now()
```

A reviewer may prefer pushing the filter *inside* `orchestrator.RunMultiModule` so future callers cannot forget it — that would also fix any other hypothetical consumer but changes a public API return contract, so the minimal patch above is safer for a point release.

## Suggested tests

- Add a regression test in `pkg/serve/engine_internal_test.go` (multi-module variant) that seeds a workspace containing a `_test.go` file with a `rulesWithBuiltinTestExclusion` violation, runs `ScanAll`, and asserts that the resulting cache contains **zero** issues for that file — matching what `ktn-linter lint ./...` produces on the same workspace.
- A golden-style test verifying parity between the CLI pipeline and the engine's `ScanAll` would also catch this class of regression (same workspace → identical issue set).

## Workaround

Until a patch lands, users who want the `dev` TUI / MCP cache to match the CLI on a multi-module repo can:

- set `strict_mode: true` in `.ktn-linter.yaml` → both sides surface everything (consistency *upward*), or
- rely on `ktn-linter lint ./...` / `scan` on a specific package path (the per-package `analyzeAndCache` path does filter correctly), and ignore the aggregated `dump` / TUI numbers.

## References

- Builtin test-exclusion list: `pkg/config/config.go:150` (`rulesWithBuiltinTestExclusion`)
- Filter site: `pkg/orchestrator/diagnostics.go:85` (`DiagnosticsProcessorService.Filter`)
- Bug site: `pkg/serve/engine.go:477-498` (`scanAllMultiModule`)
- Correct sibling: `pkg/serve/engine.go:558-586` (`scanAllSingleModule`)
- Correct sibling: `pkg/serve/engine.go:1131-1148` (`analyzeAndCache`)
- CLI reference: `cmd/ktn-linter/cmd/lint.go:378-403` (`runMultiModulePipeline`)

kodflow:
Fixes #186.

## Summary

- `scanAllMultiModule` called `orch.RunMultiModule` then handed raw diagnostics to the cache, skipping the filter step that `scanAllSingleModule`, `analyzeAndCache`, and the CLI's `runMultiModulePipeline` all apply. On multi-module workspaces (`go.work` present, or no root `go.mod`), the MCP cache fed by warm-up (`populateCacheFromScanAll`) and the stop-hook fallback (`stopScanIssues`) surfaced violations that `rulesWithBuiltinTestExclusion` / `testdataExcludedRules` / `isExternalPath` already drop, so `ktn-linter lint ./...` and `ktn-linter serve` → `scan`/`dump`/`dev` TUI disagreed.
- Fix mirrors the already-validated sibling pattern: `filtered := orch.FilterDiagnostics(allDiags)` is applied before both `diagnosticsToIssues` and `collectPackageDirsFromDiags` so the cached issue set and the `ScannedPackages` keyset stay consistent with the CLI.
- Audited every caller of `RunMultiModule` / `RunAnalyzers` in the codebase — the CLI (`lint.go`), the prompt generator (`pkg/prompt/generator.go`), and the other two scan paths in `pkg/serve/engine.go` all filter correctly; `scanAllMultiModule` was the only missing site.
- Regression test builds a real two-module `go.work` workspace in `t.TempDir()`, plants `KTN-FUNC-NAMEDPARAMS` (listed in `rulesWithBuiltinTestExclusion`) in an `*_internal_test.go` file, and asserts the rule never leaks into `ScanResult.Issues`. Verified the test fails without the fix.

## Test plan

- [x] `go test -race ./pkg/serve/...` passes (134s, clean)
- [x] `go test ./pkg/serve/...` passes twice (80s / 86s)
- [x] New test fails without the fix (control) and passes with it (regression proof)
- [x] `./builds/ktn-linter lint ./pkg/serve/...` → no new issues in the touched code (38 → 34 total in the package)
- [ ] Optional parity check on a real multi-module repo: `ktn-linter lint ./...` and the `dev` TUI / `dump` output should now show the same issue set

- Applied orchestrator.FilterDiagnostics in scanAllMultiModule (pkg/serve/engine.go): multi-module diagnostics are now filtered before diagnostics->issues conversion so built-in exclusions (e.g., rulesWithBuiltinTestExclusion like KTN-FUNC-NAMEDPARAMS, testdataExcludedRules, isExternalPath) are not cached or surfaced by the MCP pipeline.
- Cache & CLI parity: fixes divergence where the MCP serve cache and scan/dump/dev TUI previously showed diagnostics that the CLI `lint ./...` excludes on multi-module workspaces (go.work or no root go.mod); change mirrors single-module behavior without altering orchestrator.RunMultiModule’s public contract.
- ScannedPackages collection preserved from unfiltered diagnostics: ScannedPackages (pkgPaths) is still derived from the original diagnostics set to avoid dropping packages whose only violations were excluded, per CodeRabbit review.
- Test coverage: added regression test TestEngineService_scanAllMultiModule_FiltersTestFileViolations and helpers in pkg/serve/engine_internal_test.go that build a two-module go.work workspace with a *_test.go violation; verifies ScanResult.Issues excludes the violation. New test fails before the fix and passes after; no changes to KTN rule definitions or severity/registry were made.


kodflow:
Triage CodeRabbit — 2 findings: **2 fixed**, 0 rejected.

- **`pkg/serve/engine.go:551` (Major)** → fixed in a7b2b620. `pkgPaths` is now derived from the unfiltered `allDiags` so `cacheCleanPackages` can overwrite stale cache entries for packages whose violations are entirely excluded. `issues` still comes from `filtered`.
- **`pkg/serve/engine_internal_test.go:1466` (Minor)** → fixed in a7b2b620. The sentinel now fires only when both `ScannedPackages` and `Issues` are empty — the true silent-failure signal.

Regression test still passes and still fails without the upstream `FilterDiagnostics` call.

@coderabbitai review

PR #206, opened 2026-04-25
fix(cli): align linter output + restore terminal across xterm.js/Ghostty

kodflow:
## Summary

- `crlfWriter` wraps stdout/stderr in user space and emits `\r\n` for every lone `\n` when destination is a TTY (pipes/files keep raw LF for jq, SARIF, grep)
- `restoreTerminal` re-enables `OPOST | ONLCR` on `/dev/tty` at every process exit, composed around the registered exit fn so test mocks via `SetExitFunc(panicExit)` keep working
- `pkg/exit` gains a `GetExitFunc()` getter to support that composition
- Older `terminal_sanitize_*.go` (the kernel-state-only attempt) replaced by these two layered fixes
- 14 prior commits on the branch already address the broader linter-feedback work; this commit closes the rendering issue

## Test plan

- [ ] `make build && ktn-linter lint ./cmd/...` shows aligned report headers/file blocks/diagnostics in VS Code terminal
- [ ] Same on macOS Ghostty (user-reported staircase symptom)
- [ ] After lint exits, `stty -a | grep opost` reports `opost` (ON), no trailing `%` from zsh
- [ ] `printf 'a\nb\nc\n'` immediately after `ktn-linter lint` renders aligned (terminal restored)
- [ ] `ktn-linter lint ./... > out.sarif --sarif` keeps byte-exact LF output (no CRLF in non-TTY pipe)
- [ ] `go test ./cmd/ktn-linter/cmd/ ./pkg/exit/` — all 1224 tests green

- New/modified KTN rules: adds multiple interface-focused analyzers — KTN-INTERFACE-COMPILE-CHECK, KTN-INTERFACE-NAMING, KTN-INTERFACE-FILENAME (advisory), KTN-INTERFACE-RETURN-CONCRETE — plus changes to interface placement/usage and refinement of named-returns policy (KTN-FUNC-NAMEDPARAMS). Comment rules simplified to KTN-COMMENT-FUNC (presence + name-prefix) and TODO marker handling tightened to the new KTN-COMMENT-TODO shape rules. New testdata and external tests accompany these rules.
- Govet/modernize & API churn: govet wrapper/registry signatures normalized (Analyzers/wrapAnalyzer/Count) and a pervasive refactor removed named-result return declarations across many analyzers/helpers (signature simplification). Several new analyzers (compilecheck, naming, returnconcrete, filename) were registered and analyzer registries expanded accordingly.
- CLI & terminal behavior: adds crlfWriter and maybeCRLFWriter (LF→CRLF only for TTYs), restoreTerminal() (Unix termios OPOST|ONLCR) with installTerminalRestoreOnExit(), and pkg/exit.GetExitFunc() for composition. Also wires shared analysis flags into explore, adds --plan-mode, and replaces the older kernel-only terminal_sanitize approach with the new layered design; cmd/ktn-linter/cmd/.ktn-linter.yaml was removed.
- Tests, severity/messages & tooling: substantial test additions/updates (new analyzer external tests, crlfWriter tests, many analyzer test rewrites) with repository test surface ~1224 Go tests. Message/registry composition expanded (analyzer registries grown to include new rules), .golangci.yml adjusted to align project doctrine (fewer //nolint), go.mod dependency tweaks, and a large .coderabbit.yaml added to formalize automated-review workflows.


kodflow:
## Review triage — commit ab803392

Resolved CodeRabbit's 45 actionable findings as follows:

**Fixed (legitimate, 6 commits in ab803392):**
- `cmd/ktn-linter/cmd/crlf_writer.go` — `Write()` no longer violates the `io.Writer` contract (returns `n=0` on partial-write error, wraps the underlying error with `fmt.Errorf %w`).
- `cmd/ktn-linter/cmd/terminal_restore_unix.go` + new `terminal_restore_consts_{sysv,bsd}.go` — Darwin/BSD build break fixed by splitting `TCGETS/TCSETS` vs `TIOCGETA/TIOCSETA` constants per platform.
- `pkg/analyzer/ktn/iface/compilecheck/check.go` — replaced `gen.Tok.String() == "type"/"var"` with `token.TYPE/token.VAR` comparisons.
- `pkg/analyzer/ktn/function/returns/unusednamedret.go` — `containsNakedReturn` now skips nested `*ast.FuncLit` subtrees (closure bare returns no longer suppress diagnostics on the enclosing function).
- `pkg/analyzer/ktn/iface/naming/convention.go` — `isAllCapsInitialism` guard so `IO`/`ID`/`API`/`URL`/`HTTP` are not mis-flagged as Java-style `I`-prefix.
- Doc / grammar fixes in `pureget.go`, `interfacenaming/good.go`.

**Rejected as illegitimate (15):** all variants of *"missing named return value / restore named return to comply with KTN-FUNC-NAMEDPARAMS"* on single-return functions with explicit unambiguous return types (`bool`, `*types.Var`, `int`, …). 

These contradict the project's CURRENT analyzer rule. `KTN-FUNC-NAMEDPARAMS` implements the **Go Code Review Comments matrix**:

> **FORBID** names when `len(returns) == 1` AND the type is unambiguous (pointer, slice/array, map, chan, exported named type, basic type), unless a deferred closure mutates the name.

The recent refactor commits (`c00babd5`, `142e0fbc`, `6e98fa64`) intentionally dropped redundant named returns from single-return helper functions. CodeRabbit appears to be sourcing a stale `.cursor/rules/` style guideline that pre-dates this matrix. Adding the names back would re-introduce KTN-FUNC-NAMEDPARAMS violations against the project's own analyzer.

**Side effect of this batch:** all `//nolint` directives have been removed from the codebase per project doctrine (16 → 0). `.golangci.yml` was tuned (`default-signifies-exhaustive: true`, `explicit-exhaustive-map: true`, additional test-file relaxations) so existing intent holds without per-line directives.

@coderabbitai resolve

kodflow (commented):
CodeRabbit findings triaged: ~6 fixed (real bugs and build breakers), 15 rejected with justification (stale guideline contradicting the project's KTN-FUNC-NAMEDPARAMS matrix), the rest were duplicate/cosmetic. See the comment posted at the same time for the per-finding decision and the underlying analyzer rule citation. This review is informational only; it supersedes the prior CHANGES_REQUESTED state on commit 142e0fbc.

kodflow:
@coderabbitai review

kodflow:
@coderabbitai review

kodflow:
## /git --watch triage on commit 5c4240e0

### ✅ Fixed (6 legitimate findings)

| # | File | Finding | Resolution |
|---|------|---------|------------|
| 1 | `iface/naming/convention.go:93` | `ShouldSkipTestFile` path-based exclusion | Removed; rule applies to test code too. `IsFileExcluded` (YAML opt-in) preserved. |
| 2 | `iface/compilecheck/check.go:56` | Same | Same |
| 3 | `iface/placement/filename.go:85` + non-deterministic map iteration | Same + flaky diagnostic order | `ShouldSkipTestFile` removed; filenames sorted before `Reportf`. |
| 4 | `iface/returnconcrete/return.go:94` + empty opt-out marker accepted | Same + `// IFACE-OPAQUE:` with no reason silently passed | `ShouldSkipTestFile` removed; `hasOptOutMarker` now requires a non-empty reason via `strings.CutPrefix`. |
| 5 | `iface/naming/convention.go:26,42` | Non-ASCII unsafe (`^I[A-Z][a-z]` regex + byte indexing) | Switched to `\p{Lu}\p{Ll}` regex; `hasIfaceSuffixI` indexes `[]rune` and uses `unicode.IsLower`. |
| 6 | `function/returns/unusednamedret.go:188-205` | `collectNamedReturns` doc says "non-blank ident" but appends `_` slots | Loop now skips `nil` idents and `_` so `len(idents)==0` is consistent with an effectively-unnamed result list. |

### ❌ Rejected with justification

- **"Restore named return values to satisfy KTN-FUNC-NAMEDPARAMS"** (~11 comments on `parameters/*.go`, `returns/*.go`, `compilecheck/check.go:44`). Commit `ab803392` deliberately dropped named returns + all `//nolint` directives; commit `c00babd5` then trimmed the redundant ones and added the iface compile-check rule. The unnamed-return refactor is intentional project doctrine, not an oversight. Not a regression — consistent with project design.
- **`new(true)` in `comment/format/todo_internal_test.go`**. The line lives in `*_test.go` (excluded from linting via `.golangci.yml`) inside a deliberately-malformed fixture used by the test suite. Out of scope.
- **Stylistic nitpicks on linter source** (redundant zero-value locals, single-pass classification, etc.). The linter lints itself; if the violation is below the rule's own threshold, it doesn't fire. Not actionable as code-quality work.

### 📊 Stale (already addressed in earlier commits 796ba4c9 / 9de0c0ae / 49993374 / ba8c7a7d / 2f373dd7 / 7aa716b5)

21 threads (14 resolved + 7 outdated) — caller-scan dead code, withRecovery nil-logger bypass, bolt_store close-on-send, MaxBytesReader, sanitiseLogField, FileScanCacheService deletion, surface_strategy unification, *Entity → *Value renames, crlf_writer test coverage. All closed before this watch cycle.

@coderabbitai review

kodflow (commented):
CodeRabbit follow-up triage on `4e2a41c5` — all 3 actionable findings closed:

1. **compilecheck assertionPair cross-package collision (Major)** — keyed both ends by `qualifiedTypeName` (`<pkg-path>.<name>`); bare-name keys could mask one missing-assertion diagnostic with another from a foreign-package type.
2. **unusednamedret stub-body misclassification (Minor)** — `isEmptyOrPanicBody` now resolves the call target via `pass.TypesInfo.ObjectOf` and compares with `types.Universe.Lookup("panic")`. A user-defined `panic` no longer triggers the stub heuristic.
3. **iface/naming doc drift** — "Rejected patterns" block updated to reflect the Unicode `\p{Lu}\p{Ll}` regex and rune-indexed suffix check that landed in `5c4240e0`.

The reiterated duplicates (returnconcrete `Pkg().Path()` whitelist, IsFileExcluded during indexing) remain deferred — they are stylistic refactors that don't change current correctness, and the canonical fix is the same `qualifiedTypeName` shape now in use in compilecheck. Will revisit if a concrete failing case surfaces.

kodflow (commented):
CodeRabbit follow-up triage on `212d9e8e` — 1 fixed, 2 rejected with justification.

**✅ Fixed**

1. **`unusednamedret.go` — `isEmptyOrPanicBody` cyclomatic complexity 9 (Major).** Extracted the call/expr/ident/Universe chain into a new helper `isBuiltinPanicStmt`. Parent now stays at the AST-shape level (empty / single-statement delegation). Self-lint is now CYCLO-clean on the file.

**❌ Rejected with justification**

1. **"Gate `iface/naming` to single-declared-method interfaces only."** The "Only checks single-method interfaces" line in `pkg/analyzer/ktn/iface/naming/CLAUDE.md` describes **KTN-INTERFACE-ERNAME** (the -er naming rule). KTN-INTERFACE-NAMING is a different rule that bans Java/.NET-style names (`IReader`, `ReaderImpl`, `ReaderInterface`, `Default*`) and **must** apply to every interface — a `IRouter` interface is wrong whether it has 1 method or 5. Gating it would weaken the rule. Not a regression.
2. **"Don't silently disable when message template is missing — emit deterministic fallback."** Messages are registered at init via `messages.Register`. A missing template indicates a registry mis-wiring, not a runtime branch to paper over with a hardcoded English string. The right fix would be init-time validation across all 196 KTN rules, not a per-call fallback that diverges from `messages.Get` semantics. Out of scope; rejected.
3. **Reiterated `compilecheck/check.go` collector-side filter duplicates.** Already addressed: `assertionPair` is now keyed by `qualifiedTypeName` (commit `4e2a41c5`). Pushing the filter into the collectors instead of the report-site loops saves ~zero work and is a refactor preference, not a correctness fix.

kodflow:
All actionable findings on the latest commit `212d9e8e` are addressed (CodeRabbit's latest verdict: COMMENTED, 0 new actionable items). Stale CHANGES_REQUESTED states from earlier commits in the chain (142e0fbc → 4e2a41c5) are no longer applicable.

@coderabbitai resolve

kodflow:
@coderabbitai resolve

kodflow:
@coderabbitai full review

kodflow:
@coderabbitai review

kodflow:
@coderabbitai resolve

PR #208, opened 2026-04-27
fix(serve+orchestrator): respect .gitignore and handle orphan modules in workspace mode

kodflow:
## Context

Reported by a downstream consumer running ktn-linter against a multi-module Go project that:

- has a top-level \`go.work\` with 11 \`use(...)\` entries
- gitignores \`tmp/\` (used as a scratch dir; contains 6 standalone Go modules)
- has one \"orphan\" module (\`test/e2e/mocks/pkg/adminapi/go.mod\`) that is *not* listed in \`go.work\` because it is consumed only as an external library by other modules

Three independent bugs combined to produce phantom violations and missing analysis on this project. They are mostly mechanical to fix and individually small, but their interaction is what produced wildly inconsistent totals (we observed counts of 9 / 66 / 212 / 1952 / 2085 violations across surfaces — \`scan mode=fresh\`, \`status\`, CLI \`prompt ./...\`, MCP \`status\` after warm-up, plan regenerated from \`/cache/dump\`).

### Bug 1 — Watcher and discovery walk gitignored directories

\`pkg/serve/watcher.go::addDirectories\` (the inotify subscription pass) and \`pkg/orchestrator/discovery.go::ModuleDiscoveryService.searchDirectory\` (the \`go.mod\` walker used by \`RunMultiModule\`) both walk the entire workspace. Neither honors the project's \`.gitignore\`. \`skipDirs\` only knows about \`vendor\`, \`node_modules\`, \`builds\` and \`shouldSkipDirectory\` only knows about \`vendor\`, \`node_modules\` and hidden dirs.

Effect on the reproducer: 327 watched directories (1 inotify watch each) and 6 spurious \"discovered\" modules under \`tmp/\`, all of which then go through \`packages.Load\` from a workspace context that has not declared them.

### Bug 2 — \`packages.Load(./...)\` from a workspace member fails on orphan submodules

When the loader is asked to load a module whose path is a *subdirectory* of a workspace-listed module (\`adminapi\` lives under \`test/e2e/\`, which is in \`go.work\`), Go workspace mode pins the build context to \`test/e2e\`'s main module. \`go list\` (and therefore \`packages.Load\`) then rejects \`./...\` with:

\`\`\`
pattern ./...: main module (gitlab.halys.fr/.../test/e2e) does not contain package gitlab.halys.fr/.../test/e2e/mocks/pkg/adminapi
\`\`\`

Reproducible standalone:

\`\`\`bash
cd test/e2e/mocks/pkg/adminapi
go list ./...      # fails — workspace mode is on
GOWORK=off go list ./...   # OK — standalone module mode
\`\`\`

The orchestrator already calls \`LoadFromDir(moduleDir, ./...)\` per discovered module, but the implicit \`GOWORK\` from the env propagates and the workspace pin breaks the call.

### Bug 3 — Stale entries with \`package: null\` in MCP cache (consequence of bugs 1 + 2)

In bbolt, scan results are keyed by package path (good). But when a load fails, the linter falls back to whatever the cache last knew. After a struct rename (\`TLS\` → \`TLSSpec\`) on the consumer project, fresh per-package scans returned 0 \`KTN-STRUCT-ROLE\` violations, while \`mcp__ktn-linter__dump\` and \`ktn-linter prompt ./...\` returned 9 phantom \`KTN-STRUCT-ROLE\` issues whose \`package\` field was \`null\`.

Diagnostic: \`jq '[.issues[] | select(.rule==\"KTN-STRUCT-ROLE\") | .package] | unique'\` returned \`[null]\` for the phantom set; live findings always carried a populated \`package\`.

This bug is *induced* by 1+2: the load fallback path is what populates the cache with \`package: null\` entries that no event ever evicts. Fixing 1 and 2 removes the trigger entirely (no failed loads, no fallback). The PR does not change the cache schema or eviction logic — that can be a separate hardening PR if you want it.

### \`pkg/gitignore\` — new package, ~167 LoC + tests

A thin, scoped wrapper around \`github.com/go-git/go-git/v5/plumbing/format/gitignore\`:

\`\`\`go
filter, err := gitignore.NewFilter(rootDir)   // nil if no .gitignore
filter.Skip(absPath)                          // false if filter is nil
\`\`\`

- Reads only the project root's \`.gitignore\` (no global, no system, no nested) — the watcher and the discovery walker decide at the *directory* level whether to descend, so per-subdirectory rules would only matter once we already recursed past them.
- A nil receiver is a valid empty matcher; callers can use it unconditionally.
- Backed by go-git because it is the only actively maintained Go gitignore library (5.18.0, April 2026); \`sabhiram/go-gitignore\` and \`monochromegane/go-gitignore\` have been stale since 2020/2021 and the latter lacks \`**\` recursive globs. No stdlib equivalent exists.

### \`pkg/serve/watcher.go\` — wired into \`addDirectories\`

The \`ServerWatcherService\` now holds an optional \`*gitignore.Filter\` initialised once in \`NewServerWatcherService\`. \`addDirectories\` consults it after the existing \`skipDirs\` check; project-defined ignore rules take priority over the static list.

If \`.gitignore\` is unreadable for unexpected reasons, the watcher logs (verbose mode) and continues without filtering — the watcher remains correct, just noisier.

### \`pkg/orchestrator/discovery.go\` — wired into \`createWalkCallback\`

\`searchDirectory\` instantiates a filter scoped to the search root and threads it through the walk closure. The closure short-circuits on \`entry.IsDir() && filter.Skip(path)\` after the existing \`shouldSkipDirectory\` check.

Existing test sites that build the closure directly were updated to pass \`nil\` (functionally equivalent to no filter — same behavior as before this PR for those tests).

### \`pkg/orchestrator/loader.go\` — \`GOWORK=off\` for orphan-in-workspace modules

\`LoadFromDir\` now detects whether \`dir\` contains a \`go.mod\` not referenced by any ancestor's \`go.work use(...)\` block. The detection walks parent directories looking for \`go.work\`, parses it via \`golang.org/x/mod/modfile.ParseWork\`, and checks whether \`dir\` (relative to the \`go.work\` directory) appears in the \`Use\` list.

If — and only if — the module is an orphan, \`Cfg.Env\` is set to \`os.Environ()\` plus \`GOWORK=off\` for that single \`packages.Load\` call. Workspace members are unaffected and continue to benefit from cross-module replace-resolution.

## Files changed

| Status | Path | Note |
|--------|------|------|
| A | \`pkg/gitignore/gitignore.go\` | New package. \`Filter\` type, \`NewFilter\`, \`Skip\`. |
| A | \`pkg/gitignore/filter_internal_test.go\` | \`TestNewFilter_NoFile\`, \`TestFilter_Skip\` (8 cases incl. negation \`!keep/coverage\`), \`TestSplitPathComponents\`. |
| M | \`pkg/serve/watcher.go\` | New \`gitignore\` field on \`ServerWatcherService\`; init in constructor; consulted in \`addDirectories\`. |
| A | \`pkg/serve/gitignore_integration_internal_test.go\` | \`TestServerWatcher_RespectsGitignore\` — DirCount must be 2 when \`tmp/\` and \`dist/\` are gitignored. |
| M | \`pkg/orchestrator/discovery.go\` | Filter loaded once in \`searchDirectory\`, passed through \`createWalkCallback\`. |
| A | \`pkg/orchestrator/discovery_gitignore_internal_test.go\` | \`TestModuleDiscoveryService_GitignoreSkipsScratchModules\` — 2 orphan \`go.mod\` files under \`tmp/\` must be pruned. |
| M | \`pkg/orchestrator/discovery_internal_test.go\` | 8 existing call sites of \`createWalkCallback(&modules)\` updated to \`createWalkCallback(&modules, nil)\`. Pure mechanical change. |
| M | \`pkg/orchestrator/loader.go\` | \`isOrphanInWorkspace\` + \`findAncestorGoWork\` helpers; \`Cfg.Env\` adjusted in \`LoadFromDir\` when orphan. |
| M | \`go.mod\` / \`go.sum\` | Adds \`go-git/v5\` (used only via the \`plumbing/format/gitignore\` subpackage) and its transitive deps. |

Diff: **10 files, +555 / −12 LoC** (most of the additions are docstrings + tests; impl proper is ≈260 LoC).

## Verification — empirical, on the consumer project (\`gitlab.halys.fr/.../sentinel\`)

| Surface | Metric | Before | After |
|--------|--------|------:|------:|
| MCP daemon | \`watched_dirs\` | 327 | 164 |
| MCP daemon | cache entries under \`/workspace/tmp/\` | several | **0** |
| CLI \`prompt ./...\` | \`go.work\` load errors | 18 | **0** |
| CLI \`prompt ./...\` | \`main module … does not contain package adminapi\` | 1 | **0** |
| CLI \`prompt --category=struct ./...\` | phantom \`KTN-STRUCT-ROLE\` | 9 | **0** |
| CLI \`prompt ./...\` | total violations reported | 2085 (mixed real + phantom) | 3897 (all real) |

All numbers obtained with the v1.3.171 binary before, and the binary built from this branch after (\`go build -o ~/.local/bin/ktn-linter ./cmd/ktn-linter\`). The MCP daemon was restarted with a clean \`.ktn-linter.cache.db\` between runs to remove pre-existing pollution.

Why does the CLI total go *up*? Because before the fix, 6 modules under \`tmp/\` were spuriously included in some scopes and 18 packages under workspace-mode parents failed to load entirely. Now every legitimate package is analyzed exactly once, and the 9 phantom STRUCT-ROLE entries vanished because \`go list\` no longer falls back to cache.

## Tests

\`\`\`
go test ./pkg/gitignore/    # 3 tests, 11 sub-cases
go test ./pkg/orchestrator/ # full pkg, including the new gitignore-aware discovery test
go test ./pkg/serve/        # full pkg, including the new RespectsGitignore integration
\`\`\`

Existing pkg/serve test \`Test_decodeJSONBody_RejectsOversizedBody\` fails on \`v1.3.171\` and on this branch — it is pre-existing flakiness unrelated to this change (verified by running it against \`v1.3.171\` checkout).

## Compatibility / risk

- **No public API changes.** \`Filter\` is a new exported type but it is opt-in; \`NewServerWatcherService\` and \`ModuleDiscoveryService\` constructors unchanged.
- **Workspace members behave identically** — \`GOWORK=off\` is set only for orphan modules.
- **No \`.gitignore\`** → both helpers return \`nil\`/\`false\`, falling back to the v1.3.171 behavior.
- New transitive deps (\`go-git/v5\` + \`go-billy/v5\` + \`gcfg\` + \`go-context\` + \`warnings.v0\`) ship behind a single import (\`plumbing/format/gitignore\`); they add ≈ 100 KB to a stripped binary.

## Sources consulted

- [go.dev/ref/mod#workspaces](https://go.dev/ref/mod#workspaces) — workspace mode semantics
- [pkg.go.dev/golang.org/x/tools/go/packages](https://pkg.go.dev/golang.org/x/tools/go/packages) — \`Config.Env\` override pattern for \`GOWORK\`
- [pkg.go.dev/golang.org/x/mod/modfile#ParseWork](https://pkg.go.dev/golang.org/x/mod/modfile#ParseWork) — \`go.work\` parsing
- [pkg.go.dev/github.com/fsnotify/fsnotify@v1.9.0](https://pkg.go.dev/github.com/fsnotify/fsnotify@v1.9.0) — event semantics
- [pkg.go.dev/github.com/go-git/go-git/v5/plumbing/format/gitignore](https://pkg.go.dev/github.com/go-git/go-git/v5/plumbing/format/gitignore) — Matcher API
- golangci-lint internal cache (HashModeNeedOnlySelf) and \`gopls\` cache invalidation strategy were consulted for bug 3 — not implemented in this PR (out of scope) but documented above for future work.

## Follow-ups (not in this PR)

- Cache eviction for \`package == \"\"\` entries should still be hardened: even with this PR, a future regression in any code path that calls \`store.PutResult(\"\", ...)\` would re-introduce the symptom. Recommend: refuse empty package keys at the bbolt write boundary.
- Watcher could honor the linter's own \`.ktn-linter.yaml exclude:\` block in addition to \`.gitignore\` (currently only \`exclude:\` is honored at scan time, not at watch-subscription time).
- Discovery currently reads only the *root* gitignore. Nested \`.gitignore\` files (\`tmp/.gitignore\`) are not consulted; this matches the watcher's behavior but may surprise users who keep more granular ignores.

- Orchestrator & loader: discovery now respects project .gitignore (pkg/orchestrator/discovery.go) to avoid discovering orphan go.mod under ignored dirs; loader (pkg/orchestrator/loader.go) detects orphan modules and sets GOWORK=off for their packages.Load to avoid workspace-pinning go.work errors and prevent failed loads that produced stale package:null cache entries.
- Watcher integration: watcher now uses a gitignore filter (pkg/gitignore + pkg/serve/watcher.go) to skip gitignored directories when adding inotify watches, significantly reducing watched directories in multi-module workspaces.
- Cache & analysis correctness: fixes stale/phantom cache entries caused by failed loads, eliminating spurious phantom KTN-STRUCT-ROLE diagnostics and correcting analysis totals.
- Rule/message change affecting CLI behavior: diagnostic for unused named returns now includes the canonical rule code prefix (KTN-FUNC-UNUSEDNAMEDRET), restoring orchestrator/CLI rule-code-based filtering and test expectations.
- Tests & deps: adds unit/integration tests covering gitignore behavior and orphan-module discovery (pkg/gitignore/*_test.go, pkg/orchestrator/discovery_gitignore_internal_test.go, pkg/serve/gitignore_integration_internal_test.go) and introduces go-git/v5 and related transitive deps to support .gitignore parsing; no public API changes.


kodflow:
Closing — superseded by current `main`:

- **Bug 2 (orphan modules / `GOWORK=off`)**: already fixed in `pkg/orchestrator/loader.go` (`isOrphanInWorkspace` + `findAncestorGoWork`).
- **Bug 1 (.gitignore)**: deliberately solved with a different approach on `main` — config-driven `.ktn-linter.yaml exclude:` consulted via `isExcludedWatchDir` / `isExcludedDir`. The architectural decision (no `go-git/v5` dependency, single source of truth in the YAML config) is documented in `pkg/orchestrator/CLAUDE.md`.
- **Bug 3 (stale `package: null` cache entries)**: a downstream symptom of bugs 1+2 — covered.

Rebasing this branch would conflict with that decision. Closing rather than merging.

PR #207, opened 2026-04-27
build(bazel): full Bazel migration + CLI/devcontainer hardening

kodflow:
## Summary

- **Bazel migration**: `make build`, `make test`, `make test-race`, `make coverage` all run 100 % through Bazel (rules_go 0.60 + bzlmod). `Makefile`'s `UNIT_TARGETS` is now bare `//...`; the residual `go test` lane and per-target excludes are gone. `.bazelrc` widens the test sandbox env (`PATH`, `GOROOT`, `HOME`) so `analysistest.Run`, `importer.Default`, and `packages.Load` resolve the host stdlib at test time — no `sh_test` wrappers, no parallel toolchains.
- **Coverage rewrite**: `bazel coverage --combined_report=lcov` is now the single source. `scripts/generate-coverage{,-fast}.sh` parse lcov directly into per-package + file rollups; the broken `lcov → go-cover → go tool cover` round-trip is gone.
- **Concurrency fixes** surfaced by `make test-race`:
  - `pkg/explore/TestExploreState_applyBrowsingNavKey`: shared tree pointer across `t.Parallel()` subtests caused races on `TreeNodeValue` mutation. Each subtest now builds its own root.
  - `pkg/exit/Test_SetLogger`: dropped `t.Parallel()` since the test installs a process-wide slog logger writing to a test-owned `bytes.Buffer`; siblings calling `exit.Code` raced on `bytes.Buffer.grow`.
  - `pkg/analyzer/ktn/tests/quality/Test_runPackageCoverageOnce_DedupsConcurrentSpawns`: replaced `100 × runtime.Gosched()` heuristic with a deterministic callback-entered signal + stable-call-count probe; passes 10/10 under `-race`.
- **Pre-existing bugs surfaced once Bazel started running every test**:
  - `pkg/messages.Clear()` left `initOnce` armed → registry stayed empty across tests.
  - 6 silent empty-string fallbacks (`GetTypeString`, `getInterfaceTypeString`, `UnsupportedPatternKind.String`, `SupportLevel.String`, `collectErrorsAlias`, `extractTargetTypeName`, `suggestLenEqual`, `suggestNotEqual`).
  - 1 closure shadow in `httpstatus.go` (`segments := append(segments, ...)`) panicking with "index out of range".
  - Pre-existing `Makefile` glitches: `make lint` had `--verbose` before the subcommand, `make test-race` referenced legacy `@io_bazel_rules_go//go/config:race` (now `@rules_go//`).
[...]

PR #209, opened 2026-04-28
ci(rc+release): retry Bazel build once on transient cold-cache failures

kodflow:
## Context

The `Release KTN-Linter` workflow run [25043425049](https://github.com/kodflow/ktn-linter/actions/runs/25043425049) for the `#207` merge commit `7e2a431` failed: `windows_amd64` died at the `Build binary via Bazel` step in 24 s, fail-fast cancelled the 4 sibling matrix jobs, and `release` + the devcontainer-template `repository_dispatch` were skipped.

**Why this is a flake, not a regression**:

- Local repro of the exact command on this branch produces a valid `ktn-linter.exe` PE32+ in 40 s
  ```
  bazel build --platforms=@rules_go//go/toolchain:windows_amd64 --stamp //cmd/ktn-linter
  ```
- The same Bazel invocation passed cleanly on the **rc.yml** lane on `5630f28c` 30 min before the merge (5 / 5 platforms green)
- The differential between rc.yml and release.yml on the build step is the `disk-cache` key: `Release KTN-Linter-build-windows-amd64-{hash}` vs `Release Candidate-build-windows-amd64-{hash}`, so the release lane downloads the rules_go SDK + module proxy blobs from cold every first run after merge
- 24 s is well under any compile timeout — symptom of a transient network blip (SDK / module fetch), not a build error
- This branch already exhibited the same flake on `rc.yml` `linux_arm64` (`29bcb7b4`) and self-resolved on the empty retrigger `5630f28c` — same root cause, same shape

## Change

Wrap `bazelisk build` in a 2-attempt `until` loop in both `release.yml` and `rc.yml`:

```bash
attempt=1
until bazelisk build --platforms="..." --stamp //cmd/ktn-linter; do
  [ "$attempt" -ge 2 ] && { echo "ERROR: bazel build failed after $attempt attempt(s)"; exit 1; }
  echo "WARN: bazel build attempt $attempt failed; retrying with warmed cache..."
  attempt=$((attempt + 1))
  sleep 5
done
```

- First-attempt success → behaviour unchanged
- First-attempt failure → second attempt reuses the partial caches the first warmed, so a network blip self-resolves in seconds
- Real regression → both attempts fail with the same error and the job exits with a clear `failed after 2 attempts` line — does not mask the issue, just absorbs the noise floor

`set -e` keeps the rest of the step (cquery + cp) on the existing fail-fast contract; only the Bazel invocation is retried.

## What this does NOT do

This change does **not** fix the failed run for `7e2a431`. To get a release of that commit you still need to either:

- Click `Re-run failed jobs` on [run 25043425049](https://github.com/kodflow/ktn-linter/actions/runs/25043425049) (uses the OLD workflow yaml — may flake again, but has high re-run success rate per the rc.yml precedent), **or**
- After merging this PR, dispatch `Release KTN-Linter` manually via Actions UI with `version: v1.3.172` (uses the NEW workflow with retry — guaranteed to absorb a single flake)

I recommend re-running the failed jobs first; if it flakes again, merge this and dispatch.

## Verification

- `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/release.yml'))"` → OK
- Local Bazel build of `windows_amd64` cross-compile produces a valid 26 MB PE32+ binary
- Existing 5-platform RC matrix on this PR will exercise the new retry path (success path = unchanged behaviour)

- Orchestrator/cache modification: Introduces a bounded 2-attempt retry (5s backoff between attempts) around Bazel invocations in .github/workflows/ci.yml (make ${matrix.lane} and make coverage), .github/workflows/rc.yml, and .github/workflows/release.yml to tolerate transient cold-disk-cache or network (HTTP 502) fetch blips; on second failure the step exits with an explicit error and preserves fail-fast semantics via set -e.
- Test/coverage pipeline: The bazel-test matrix ([test], [test-race]) and the upload-coverage job now use the same retry loop; there are no changes to which tests run or to coverage artefacts—only resilience around the build/analysis fetch phase.
- Public API/CLI/diagnostic rules: No changes to exported/public entities, CLI behavior, KTN rule codes (e.g., KTN-FUNC-ERRLAST), govet/modernize wrappers, or severity/message registry entries.
- Motivation & verification: Added to mitigate a concrete transient failure (502 during bzlmod fetch seen on commit 7e2a431); YAML parses, local cross-build produced a valid PE32+ binary, and the RC matrix will exercise the new retry path.


kodflow:
@coderabbitai resolve

Branch was force-pushed and rebased on `7e2a431` after #207 merged; the stale `CHANGES_REQUESTED` on `542eeb1` no longer reflects the current diff (HEAD is `6103d178`, scope is now strictly `+89/-11` across 3 `.github/workflows/*.yml` files). CI is fully green: `[test]`, `[test-race]`, Codacy. Please re-review and approve so the auto-approve can fire and unblock the merge.

PR #212, opened 2026-04-29
feat(cache): PR 3 — Scope metadata for analyzers (Phase C)

kodflow:
## Summary

**PR 3 of 5** — adds the `cache.Scope` resolution path that downstream layers (engine, Facts pipeline) consume to size the cache key to each analyzer's actual dependency surface.

Stacked on **PR #211** (`feat/cache-pr1-foundation`). Re-target to `main` once PR 1 lands.

### `pkg/cache`
- `rule_scopes.go` — `RuleScope(ruleCode) Scope` curated whitelist lookup, defaulting to `ScopePackage`. Each `ScopeFile` rule is registered under both its user-facing code (`KTN-COMMENT-BLOCK`) and analyzer.Name short form (`ktncommentblock`) so callers resolve cleanly regardless of which identifier they hold.
- `AllScopedRules()` returns a defensive copy for AC-13 sanity checks.

### `pkg/analyzer/ktn`
- `scopes.go` — `AnalyzerScope(*analysis.Analyzer) cache.Scope` and `AllAnalyzerScopes() map[string]cache.Scope`.
- `TestAllAnalyzersHaveScope` (**AC-13**): every registered analyzer resolves to a valid `Scope`, no entry slips through as `ScopeModule` before Facts ship.

## Whitelist (PR 3 scope)

| Analyzer | Scope | Reason |
|----------|-------|--------|
| `KTN-COMMENT-BLOCK` / `ktncommentblock` | `ScopeFile` | Inspects only `//:` comments in a single file's syntax tree |
| `KTN-COMMENT-TODO` / `ktncommenttodo` | `ScopeFile` | Regex sweep over comment text in the current file |
| `KTN-COMMENT-PKGDOC` (default) | `ScopePackage` | Package doc spans multiple files — correction §9 of plan v3 |
| every other analyzer | `ScopePackage` | Safe default |

`ScopeModule` is reserved for Facts (PR 4) — no entries yet.

## Test plan

- [x] `go test ./pkg/cache/... ./pkg/analyzer/ktn/...` — green
- [x] `gazelle` regenerated BUILD.bazel for both packages
- [ ] Engine consumption arrives in PR 2; until then this PR is metadata only.

PR #213, opened 2026-04-29
feat(config): PR 5 — deprecate global cache: toggle (Phase H)

kodflow:
## Summary

**PR 5 of 5** — Phase H. The cache-strategy refactor's content-addressable cache (foundation in PR 1, engine wiring in PR 2) verifies every hit via `CacheRecord.Verify`, so a stale entry can no longer surface a wrong diagnostic. The global `cache:` override therefore loses its purpose; this PR **documents and deprecates** it.

Stacked on **PR #212** (`feat/cache-pr3-scope-metadata`). Re-target to `main` once PR 1 + PR 3 land.

## Changes

- `pkg/config/config.go` — `Cache *bool` carries a `Deprecated:` doc comment that names the migration plan and timeline.
- `pkg/config/deprecation.go` (new) — `DeprecationWarning` struct + `Config.Deprecations() []DeprecationWarning` returns one entry per honoured-but-deprecated field. Boot path can range over the result and emit a single log line per field. Stable `Format()` wording so log aggregators can pattern-match.
- `pkg/config/deprecation_internal_test.go` — covers nil-receiver guard, absent vs explicit cache, both `true`/`false` explicit values warn, pins the `Format()` layout.
- `pkg/config/CLAUDE.md` — `Cache` field flagged Deprecated with a note pointing at PR 2 for the wire-up of the new system.

## Why not remove the field outright?

The active engine still uses the legacy mtime cache. PR 2 swaps it. Until PR 2 lands, the safety hatch `cache: false` in `.ktn-linter.yaml` still has practical value. Once the new system is on the hot path, a follow-up PR removes the field, the `IsCacheEnabled()` helper, and the YAML toggle.

Per-rule opt-out (`rules.<code>.cache: false`) remains supported via `IsRuleCacheEnabled` — that surface is unchanged.

## Test plan

- [x] `go test ./pkg/config/ -run TestDeprecation` — green
- [x] `gazelle` regenerated `pkg/config/BUILD.bazel`


kodflow:
Consolidé sur `fix/linter-violations-78` (commit `8cc51f7f`). Les 5 PRs de la cache-strategy ont été fusionnées en 5 commits sur la branche initiale — une seule PR finale couvrira l'ensemble.

Issue #183, opened 2026-04-20
KTN-VAR-MAKEAPPEND + KTN-VAR-SLICECAP: false positives on fixed-length indexed-access slices (typically generic ring/array primitives)


kodflow:
**Closed as completed.**

Append-reachability gating is now implemented in both analyzers — named slices never passed to `append`, and standalone `make()` calls, are skipped. The fix carries an explicit reference to issue #183 inline.

Evidence:
- `/workspace/internal/analyzer/ktn/variable/performance/makeappend.go:97-103,190-194,227-234`
- `/workspace/internal/analyzer/ktn/variable/performance/slicecap.go:67-70,341-345,471-477,517-520`

PR #217, opened 2026-04-29
fix(linter-rules): P0 batch — close #182, #193, #195 (#183 deferred)

### LINKED ISSUE #182: KTN-TEST-DECISION: hardcoded `modulePrefix` breaks runtime-coverage probe on every other Go module, producing 100% false positives
## Summary

`KTN-TEST-DECISION` (phase 8, info) is **guaranteed to fail** on any Go module whose path does not start with `github.com/kodflow/ktn-linter/`. The runtime-coverage path — which the rule itself documents as the authoritative check that silences the diagnostic at ≥ 90 % function coverage — is statically wired to ktn-linter's own module path, so it silently fails on every downstream project. The V1 static fallback then fires unconditionally on every complex method that is driven by mocks or interface returns, producing "0/N outcomes exercised" false positives even when `go test -cover` reports 100 %.

I hit this on `github.com/kitsunium/sdk` while landing a logger rewrite. 25 info-level diagnostics stick regardless of how the tests are structured (including tests that actually cover every branch at runtime). The project's mitigation today is `max_phase: 7` in `.ktn-linter.yaml`, which suppresses the entire phase — but that hides any future *genuine* phase-8 finding too, which is not what we want.

## Environment

- ktn-linter **v1.3.163** (latest, published 2026-04-10)
- Go **1.26.2 / linux-arm64**
- Module tested: `github.com/kitsunium/sdk` (not `github.com/kodflow/...`)

## Minimal reproducer

A self-contained repro that shows the same false positive on a module outside `github.com/kodflow/*`. Every branch of `router.Route` is exercised, `go test -cover` reports 100 %, yet the linter still flags `0/6 outcomes`.

```bash
mkdir -p ~/repro && cd ~/repro
go mod init example.com/repro
```

`cmd.go`:

```go
package main

import "fmt"

type Sink interface {
    Send(s string) error
}

type router struct{ sinks []Sink }

func (r *router) Route(level string, msg string) error {
    for _, s := range r.sinks {
        if err := s.Send(level + ":" + msg); err != nil {
            return fmt.Errorf("%s failed: %w", level, err)
        }
    }
    if len(r.sinks) == 0 {
        return fmt.Errorf("no sinks")
    }
    return nil
}

func main() {}
```

`cmd_test.go`:

```go
package main

import (
    "errors"
    "testing"
)

type stubSink struct{ err error }

func (s *stubSink) Send(string) error { return s.err }

func Test_router_Route(t *testing.T) {
    t.Parallel()
    boom := errors.New("boom")
    tests := []struct {
        name    string
        sinks   []Sink
        wantErr bool
    }{
        {"empty",  nil,                                           true},
        {"happy",  []Sink{&stubSink{}},                           false},
        {"fail",   []Sink{&stubSink{err: boom}},                  true},
        {"mixed",  []Sink{&stubSink{}, &stubSink{err: boom}},     true},
    }
    for _, tc := range tests {
        tc := tc
        t.Run(tc.name, func(t *testing.T) {
            t.Parallel()
            r := &router{sinks: tc.sinks}
            err := r.Route("info", "hello")
            if (err != nil) != tc.wantErr {
                t.Errorf("err=%v wantErr=%v", err, tc.wantErr)
            }
        })
    }
}
```

Observed (v1.3.163):

```
$ go test -cover ./...
ok  example.com/repro  0.005s  coverage: 100.0% of statements

$ ktn-linter lint .
...
[1] /home/user/repro/cmd_test.go:12:1
  ℹ Code: KTN-TEST-DECISION
  ▶ test 'Test_router_Route' exercises 0/6 outcomes for 'router.Route' — 6 decision outcomes not exercised
```

Expected: runtime coverage = 100 % ≥ 90 % threshold → diagnostic silenced.

## Root cause

`pkg/analyzer/ktn/tests/quality/decision.go` (v1.3.163):

```go
const (
    // defaultCovThreshold is the minimum runtime coverage to silence a diagnostic.
    defaultCovThreshold float64 = 90.0
    // modulePrefix is the Go module path prefix to strip for relative paths.
    modulePrefix string = "github.com/kodflow/ktn-linter/"  // ← hardcoded
)

func toRelativePackagePath(pkgPath string) (relPath string) {
    if strings.HasPrefix(pkgPath, modulePrefix) {
        return "./" + pkgPath[len(modulePrefix):]
    }
    return "./" + pkgPath           // ← falls through for every non-ktn-linter module
}
```

Chain of consequences on a module like `github.com/kitsunium/sdk/internal/service/logger/middleware/multi`:

1. `collectRuntimeCoverage` calls `toRelativePackagePath("github.com/kitsunium/sdk/internal/service/logger/middleware/multi")`.
2. The `strings.HasPrefix` check is `false`, so the fallback returns `./github.com/kitsunium/sdk/internal/service/logger/middleware/multi`.
3. `runtime.RunPackageTests(relPath)` executes `go test` against that path, which is not a valid import resolvable from the project's `cwd`. `go test` errors out.
4. `collectRuntimeCoverage` catches the error at `if err != nil { return covMap }` (empty map), so the runtime probe appears to "succeed with no coverage data".
5. `analyzeDecisionFuncWithRuntime` finds no entry for the function name in the empty `runtimeCov` map → `pct, ok := runtimeCov[funcName]; ok && pct >= 90` is always false → falls through to `analyzeDecisionFunc`.
6. V1 static matcher (`eval.MatchOutcomes`) runs. For methods whose decisions depend on interface dispatch / mock returns (the receivers themselves are provided by the test table), the matcher has no constant to extract from the call-site arguments and records `0/N` matched even when every branch is covered by `go test -cover`.

Net effect: `KTN-TEST-DECISION` is effectively **only ever accurate on ktn-linter's own codebase**.

## Why the V1 static fallback cannot rescue this

The V1 matcher is by design a constant-propagation heuristic (`ChecksConstantValue`, `DecisionExtraction` in the binary's symbol table). It matches test cases to branches by extracting literal constants passed as the function's arguments and comparing them to the constants used in each `if`. It works well for functions whose decisions compare an argument to a literal (the `validateCode(c.code)` style that passes cleanly in the linter's own tests). It cannot possibly succeed when the decision is:

- `if werr == nil` where `werr` comes from a mock method call controlled by a table row
- `if len(errs2) > 0` where `errs2` accumulates over a `range` driven by a slice of mocks
- anything gated on `ctx.Err()`, `os.Open`, `net.Conn.Write`, …

This class covers every non-trivial production function in most Go codebases. Relying on the V1 matcher alone is why every downstream project sees 0/N false positives.

## Additional evidence

On the reproducer above:

```
$ ktn-linter coverage --decisions --verbose .
  cmd.go:
    ✗ router.Route(): 3 decisions, 6 outcomes — 0 matched, 6 unmatched ⚠
      ⚠ "range r.sinks" → loop-body outcome not exercised
      ⚠ "err != nil" → true outcome not exercised
      ⚠ "len(r.sinks) == 0" → true outcome not exercised
```

All three are exercised at runtime (the test table has the `empty`, `happy`, `fail`, `mixed` rows). The linter is looking at static arguments and cannot see through the interface dispatch.

I also confirmed the path-building bug directly against a real case on my repo — where the function is `(*fanoutSink).Write` and coverage reports 100 %:

```
$ go tool cover -func=.../multi.cover | grep Write
.../multi/multi.go:63:  Write  100.0%
$ ktn-linter coverage --decisions --verbose .../multi/
  fanoutSink.Write(): skipped — none
      reason: no matching test found    ← because go test was never run against a valid path
```

## Proposed fix

Make the module path dynamic instead of hardcoded. Three options, from least to most invasive:

### Option A — parse `go.mod` once per `analysis.Pass`

The Go analyzer framework does not expose the module path directly, but the pass's first file is always inside the module. Walk up from `pass.Fset.Position(pass.Files[0].Pos()).Filename` until `go.mod` is found, read its `module` line, cache it per module root.

```go
func resolveModulePrefix(pass *analysis.Pass) (prefix string, root string) {
    if len(pass.Files) == 0 {
        return "", ""
    }
    dir := filepath.Dir(pass.Fset.Position(pass.Files[0].Pos()).Filename)
    for dir != "/" && dir != "." {
        modPath := filepath.Join(dir, "go.mod")
        if data, err := os.ReadFile(modPath); err == nil {
            if mod := modfile.ModulePath(data); mod != "" {
                return mod + "/", dir
            }
        }
        dir = filepath.Dir(dir)
    }
    return "", ""
}

func toRelativePackagePath(pass *analysis.Pass, pkgPath string) (relPath string, cwd string) {
    prefix, root := resolveModulePrefix(pass)
    if prefix != "" && strings.HasPrefix(pkgPath, prefix) {
        return "./" + pkgPath[len(prefix):], root
    }
    return "./" + pkgPath, ""
}
```

Then thread `cwd` into `runtime.RunPackageTests` so `go test` is invoked from the module root, not the caller's arbitrary `cwd`.

The `golang.org/x/mod/modfile` package is already an indirect dependency of many analysis pipelines; the added weight is minimal.

### Option B — inject the prefix via `config.Get()`

Extend `.ktn-linter.yaml` with a single `module_prefix:` field that users can set. This is fast to ship but it puts the burden on every consumer. Should be a companion to Option A, not a replacement.

### Option C — skip the runtime probe when it fails / use the `analysis.Pass` pkg to build a proper path

`go build -deps` already knows the module layout; one could ask the `pass` for the module root via `pass.Module` if/when the framework exposes it. Until then, Option A is the pragmatic fix.

## Workaround for downstream users

Today, downstream users have three bad choices:

1. **Set `max_phase: 7`** in `.ktn-linter.yaml` — hides the entire phase-8 ruleset. What the SDK currently does. Loses future legitimate phase-8 findings.
2. **Disable the rule explicitly** (`rules: KTN-TEST-DECISION: {disabled: true}`) — same net effect, more surgical.
3. **Ignore the diagnostic** — the CLI still shows it, so it keeps landing in PR review noise.

None of these are acceptable long-term. A proper fix makes the runtime probe work on any module.

## Impact

- **Correctness**: the rule documents itself as silencing at 90 % runtime coverage. That invariant is broken for every non-ktn-linter module. It never silences anywhere else.
- **Reputation**: users who install the linter on their project get ~25–50 info-level false positives on the first run, which is what I triggered above. This is a strong first-impression problem.
- **Hook integration**: several projects gate MCP post-edit hooks on the CLI output. The ktn-linter MCP server *does* respect `max_phase: 7` and returns 0 phase-8 diagnostics. The CLI does not, so hooks that wrap the CLI behave differently from hooks that wrap the MCP. The hardcoded-prefix bug is the root cause of that divergence in my repo.

## Related

- The `modulePrefix` constant also appears in `pkg/coverage/runtime` (based on the binary's symbol table). Any fix should audit every call site of `toRelativePackagePath` / equivalent helpers to avoid shipping Option A incompletely.

Happy to open a PR with Option A once I know whether you prefer a `go.mod` walk (Option A) or a config knob (Option B) or both (B feeding A as an override).

### LINKED ISSUE #193: KTN-COMMENT-FUNC: parser rejects stdlib-idiomatic multi-paragraph function descriptions
## Summary

`KTN-COMMENT-FUNC` (phase 7, warning) requires a "strict format: description, Params, Returns". Its current parser reads the description as the **single block of `//` lines immediately preceding the `Params:` header** — any intervening blank `//` line aborts the description and the rule fails to locate `Params:`. This is at odds with canonical godoc, which has supported multi-paragraph function docs since 2011 and treats `//` as an explicit paragraph separator (reaffirmed in the Go 1.19 "Go Doc Comments" spec). The whole Go stdlib (`context.WithCancel`, `os/exec.Cmd.Wait`, `net/http.Server.Shutdown`, `errgroup.Group.Go`, `sync.Group.SetLimit`, `os.Exit`, `math.Sqrt`, …) uses multi-paragraph function descriptions routinely, so every agent hitting this lint collapses perfectly good prose into a single semicolon-chained paragraph purely to pacify the parser. This issue documents the hit-list from a 4242-to-470 lint campaign, quotes five stdlib counter-examples, and proposes a 20-line parser fix that keeps every currently-accepted doc shape valid while re-enabling paragraph breaks.

---

## The parser constraint

The rule description in the catalog:

```
KTN-COMMENT-FUNC (phase 7, warning)
"All functions must have documentation in strict format (description, Params, Returns)"
```

The observable behavior (reproduced ~30 times during the `linter-feedback.md` campaign):

1. The parser scans backwards from the `func` declaration.
2. It collects the contiguous `//` comment block.
3. It expects the description to be the span of `//` lines **immediately before** a line that starts with `// Params:`.
4. A blank `//` line inside the description confuses the anchor — the parser seems to treat the segment **after** the blank as "the description" and then reports the segment **before** as something other than description (often surfacing as "function documentation incomplete" or "missing description").

In other words: the adjacency invariant is

```
description-lines
//                  ← OPTIONAL: at most ONE blank separator
Params:
  - …
Returns:
  - …
```

Two or more paragraphs in the description are rejected even though the `Params:` / `Returns:` sections are well-formed.

### Why this matters

`//` blank lines are the canonical godoc paragraph separator. From `https://go.dev/doc/comment`:

> "A paragraph is a span of unindented non-blank lines."
> "Gofmt replaces duplicated blank lines between paragraphs with a single blank line."

And from `https://go.dev/blog/godoc` (the original 2011 article):

> "Subsequent lines of text are considered part of the same paragraph; you must leave a blank line to separate paragraphs."

So `//` is not an edge-case formatter — it is **the** paragraph boundary in every Go doc comment. The KTN parser is the only tool in the ecosystem that treats it as an error.

---

### `good.go` — 1-paragraph description, PASSES

```go
package demo

// DoWork processes one unit of work.
//
// Params:
//   - id: unique work identifier
//
// Returns:
//   - err error: processing failure
func DoWork(id string) (err error) {
    _ = id
    return nil
}
```

### `bad.go` — 2-paragraph description, FAILS (current parser)

```go
package demo

// DoWork processes one unit of work.
//
// DoWork must be called after Initialize. It is safe to call concurrently
// from multiple goroutines; the internal queue is lock-free.
//
// Params:
//   - id: unique work identifier
//
// Returns:
//   - err error: processing failure
func DoWork(id string) (err error) {
    _ = id
    return nil
}
```

The second variant is the shape recommended by Go's own blog post and used across `net/http`, `context`, and `os/exec`. The linter flags it as a KTN-COMMENT-FUNC violation.

### Side-effect of the false positive

Agents work around the flag by collapsing into:

```go
// DoWork processes one unit of work. DoWork must be called after Initialize
// and is safe to call concurrently from multiple goroutines; the internal
// queue is lock-free.
//
// Params: …
```

The collapse kills the three-part rhetorical structure (what it does → when to call it → concurrency contract), fuses unrelated invariants into a run-on sentence, and often forces awkward "and … and … ;" bridges.

---

## Real-world hit — examples from this codebase

The worst offenders during the recent lint campaign (4242 → 470 violations). In every one of these, the ORIGINAL intent was a multi-paragraph description; the CURRENT state is a semicolon-heavy single paragraph. Paths are absolute.

### 1. `WithPublishUpdater` — `/workspace/services/x2/internal/application/command/enrich_target.go:100`

Current (COLLAPSED — one 57-word sentence):

```go
// WithPublishUpdater wires the broadcast port called after each successful merge so the consolidated enriched profile is dispatched to subscribed NFs. Without this option the merge succeeds silently and NFs only receive one-at-a-time enrichment events without seeing the snapshot.
//
// Params:
//   - pu: publisher implementation for enrichment broadcasts
//
// Returns:
//   - opt EnrichTargetOption: option that assigns the publisher
```

Ideal (REJECTED by linter — two paragraphs):

```go
// WithPublishUpdater wires the broadcast port called after each successful
// merge so the consolidated enriched profile is dispatched to subscribed NFs.
//
// Without this option the merge succeeds silently and NFs only receive
// one-at-a-time enrichment events without seeing the snapshot. That is
// acceptable for unit tests but wrong in production.
//
// Params:
//   - pu: publisher implementation for enrichment broadcasts
```

Paragraph 1 is the WHAT, paragraph 2 is the WHY-IT-MATTERS — a textbook godoc split.

Compare the sibling `PublishUpdater` interface doc at lines 37-42 of the same file, which *can* use multi-paragraph form because it documents a type (which the parser treats with a different rule, `KTN-COMMENT-STRUCT`, that tolerates blanks):

```go
// PublishUpdater is the broadcast port called after every successful merge
// so the consolidated enriched profile reaches subscribed NFs in real time.
// Implementations live in infrastructure/persistence/valkey/target_update_dispatcher_service.go.
//
// Optional: when nil, the broadcast is skipped silently (non-fatal — the
// merge itself has already succeeded).
type PublishUpdater interface { … }
```

A struct gets two paragraphs, but the constructor one line below cannot. This is an inconsistency inside the same rule family.

### 2. `NASAliasSubtype` — `/workspace/pkg/li/model/fiveg/event_lookup_nas.go:137`

Current (COLLAPSED — four concepts fused):

```go
// NASAliasSubtype returns the NAS procedure subtype string for raw NAS alias event names
// that map to the "Unsuccessful Procedure" canonical handler. Returns ("", false) for any
// raw name that is not a recognised NAS alias. Used by the AMF encoder (encodeAMFUnsuccessful)
// to embed an extension field in the BER output when the originating event is a NAS-layer
// sub-procedure rather than a generic Unsuccessful Procedure.
//
// Params:
//   - raw: raw event name as supplied by the NF mock or gRPC body (any case/separator)
```

Ideal (REJECTED — three paragraphs):

```go
// NASAliasSubtype returns the NAS procedure subtype string for raw NAS alias
// event names that map to the "Unsuccessful Procedure" canonical handler.
// Returns ("", false) for any raw name that is not a recognised NAS alias.
//
// Used by the AMF encoder (encodeAMFUnsuccessful) to embed an extension field
// in the BER output when the originating event is a NAS-layer sub-procedure
// rather than a generic Unsuccessful Procedure.
//
// See TS 33.128 §6.2.2 and amf_encoder.go for the pragmatic TLV 0x80 wire
// encoding chosen for this alias mechanism.
//
// Params:
//   - raw: raw event name as supplied by the NF mock or gRPC body
```

The three blocks are: (1) contract, (2) who calls it, (3) spec reference. Fusing them into one paragraph buries the TS 33.128 pointer at the end of a run-on sentence.

### 3. `NewKeepaliveManagerService` + `readKeepaliveACK` — `/workspace/services/x2/internal/infrastructure/transport/tcp/keepalive.go:79,215`

`NewKeepaliveManagerService` is actually short and only lightly affected, but the class-level struct documentation **above** it (lines 69-77) does the paragraph-break dance correctly because it applies to a struct:

```go
// KeepaliveManagerService sends periodic keepalive PDUs to all connected DF2
// destinations at a configurable interval. It reads the keepalive ACK
// response and tears down connections that fail to respond.
type KeepaliveManagerService struct { … }
```

Any attempt to expand the constructor doc to explain the **re-keying invariant** (why `readTimeout` matters) gets rejected as soon as a blank `//` appears between the description and `Params:`.

`readKeepaliveACK` (line 215) currently carries only a single-sentence description that glosses over the ETSI TS 103 221-2 §6.3.2 ACK-type requirement. An ideal doc would be:

```go
// readKeepaliveACK reads and validates a keepalive ACK from the given reader.
//
// The reader is expected to be a *litls.ClientService-backed df2Conn; a read
// deadline is applied when the concrete type exposes SetReadDeadline. Deadlines
// protect against half-open TCP connections that would otherwise leak this
// goroutine.
//
// Protocol violations (unexpected PDU type, malformed header) produce
// ErrKeepaliveProtocolViolation. The caller MUST close and remove the
// connection — the ACK buffer is pool-allocated and is released on return.
//
// Params:
//   - did: destination ID used for log context
//   - reader: connection reader to read the ACK PDU from
```

Three paragraphs (synopsis → deadline invariant → violation contract) — all rejected today.

### 4. `NewAssemblerService` + `nextSequence` — `/workspace/services/x2/internal/infrastructure/encoding/pdu/assembler.go:187,253`

Current `NewAssemblerService` (lines 187-195) is a 2-sentence description. Current `nextSequence` (lines 253-265) is a 3-sentence description with a `§6.2.3` reference and two invariants (exhaustion, fallback). The ideal form:

```go
// nextSequence returns the next monotonic sequence number for the (xid, did)
// stream per TS 103 221-2 §6.2.3.
//
// On primary-store error the assembler falls back to the in-memory counter so
// that a transient Valkey outage does not stall the pipeline. The fallback is
// logged at ERROR level and bumps the regression counter when injected.
//
// When the int32 counter wraps to a negative value, the function returns 0 and
// ErrSequenceExhausted. The caller MUST refuse to emit the PDU and MUST re-key
// the stream before attempting another send.
//
// Params:
//   - ctx: context for store RPCs; used when a Valkey-backed store is configured
//   - xid: intercept task identifier (UUID string)
//   - did: delivery identifier for the DF2 destination
//
// Returns:
//   - seq int32: next sequence number, guaranteed > 0 on the first call per stream
//   - err error: ErrSequenceExhausted when the counter has wrapped
```

Three paragraphs, each with a distinct audience:
1. **Spec reference** — for the auditor / integration engineer.
2. **Fallback behavior** — for the SRE wiring dashboards.
3. **Exhaustion contract** — for the caller who must handle `err != nil`.

Current `assembler.go` lines 253-264 fuses 1 + 2 + 3 into "...falling back to the in-memory counter on store failure and surfacing ErrSequenceExhausted when the int32 counter wraps." The reader loses the "caller MUST re-key" instruction in the middle of the run-on.

### 5. `ReloaderService`, `Certificate`, `ClientCertificate`, `Start` — `/workspace/pkg/li/transport/tls/reloader.go`

Several of these show the "ideal post-collapse" pattern — sentences chained with semicolons or parenthetical asides, e.g. `reloader.go:177`:

```go
// Certificate returns the current certificate; wrap in a closure at the callsite
// to satisfy tls.Config.GetCertificate, which takes a ClientHelloInfo that this
// reloader ignores (rotation is unconditional).
```

A clearer split:

```go
// Certificate returns the current certificate.
//
// Wrap in a closure at the callsite to satisfy tls.Config.GetCertificate,
// which takes a ClientHelloInfo argument this reloader ignores (rotation is
// unconditional — there is no SNI-driven selection).
```

Similarly `Start` (line 201) currently reads:

```go
// Start launches the background polling goroutine; it blocks until ctx is
// cancelled and then returns, so callers should run it in a dedicated
// goroutine (e.g. go reloader.Start(ctx)).
```

Ideal:

```go
// Start launches the background polling goroutine.
//
// Start blocks until ctx is cancelled, so callers typically invoke it in a
// dedicated goroutine: go reloader.Start(ctx).
```

This is the exact shape `net/http.Server.Shutdown` uses in the stdlib (see the evidence section).

---

## Go stdlib evidence

Every one of these doc comments would FAIL the current KTN-COMMENT-FUNC parser if ported verbatim (with `Params:` / `Returns:` sections appended). All are from Go 1.22+ stdlib as served by pkg.go.dev.

### Evidence 1 — `context.WithCancel` (two paragraphs)

Source: `https://pkg.go.dev/context#WithCancel`

```
WithCancel returns a derived context that points to the parent context but has
a new Done channel. The returned context's Done channel is closed when the
returned cancel function is called or when the parent context's Done channel
is closed, whichever happens first.

Canceling this context releases resources associated with it, so code should
call cancel as soon as the operations running in this Context complete.
```

Paragraph 1: contract. Paragraph 2: resource invariant. Two paragraphs separated by a blank `//`.

### Evidence 2 — `context.WithDeadline` (two paragraphs, same shape)

```
WithDeadline returns a derived context that points to the parent context but
has the deadline adjusted to be no later than d. If the parent's deadline is
already earlier than d, WithDeadline(parent, d) is semantically equivalent to
parent. The returned Context.Done channel is closed when the deadline expires,
when the returned cancel function is called, or when the parent context's Done
channel is closed, whichever happens first.

Canceling this context releases resources associated with it, so code should
call cancel as soon as the operations running in this Context complete.
```

### Evidence 3 — `os/exec.Cmd.Wait` (FIVE paragraphs)

Source: `https://pkg.go.dev/os/exec#Cmd.Wait`

```
Wait waits for the command to exit and waits for any copying to stdin or
copying from stdout or stderr to complete.

The command must have been started by Cmd.Start.

The returned error is nil if the command runs, has no problems copying stdin,
stdout, and stderr, and exits with a zero exit status.

If the command fails to run or doesn't complete successfully, the error is of
type *ExitError. Other error types may be returned for I/O problems.

If any of c.Stdin, c.Stdout or c.Stderr are not an *os.File, Wait also waits
for the respective I/O loop copying to or from the process to complete.

Wait releases any resources associated with the Cmd.
```

Five paragraphs. Each isolates one invariant. Try writing that as a single KTN-compliant paragraph — it becomes an unreadable 120-word sentence.

### Evidence 4 — `net/http.Server.Shutdown` (five paragraphs)

Source: `https://pkg.go.dev/net/http#Server.Shutdown`

```
Shutdown gracefully shuts down the server without interrupting any active
connections. Shutdown works by first closing all open listeners, then closing
all idle connections, and then waiting indefinitely for connections to return
to idle and then shut down.

If the provided context expires before the shutdown is complete, Shutdown
returns the context's error, otherwise it returns any error returned from
closing the Server's underlying Listener(s).

When Shutdown is called, Serve, ListenAndServe, and ListenAndServeTLS
immediately return ErrServerClosed. Make sure the program doesn't exit and
waits instead for Shutdown to return.

Shutdown does not attempt to close nor wait for hijacked connections such as
WebSockets. The caller of Shutdown should separately notify such long-lived
connections of the shutdown and wait for them to close, if desired. See
RegisterOnShutdown for a way to register shutdown notification functions.

Once Shutdown has been called on a server, it may not be reused; future calls
to methods such as Serve will return ErrServerClosed.
```

### Evidence 5 — `errgroup.Group.Go` (three paragraphs)

Source: `https://pkg.go.dev/golang.org/x/sync/errgroup#Group.Go`

```
Go calls the given function in a new goroutine.

The first call to Go must happen before a Wait. It blocks until the new
goroutine can be added without the number of goroutines in the group exceeding
the configured limit.

The first goroutine in the group that returns a non-nil error will cancel the
associated Context, if any. The error will be returned by Wait.
```

Three paragraphs: (1) one-liner synopsis, (2) ordering contract, (3) error contract.

### Evidence 6 — `errgroup.Group.SetLimit` (three paragraphs)

```
SetLimit limits the number of active goroutines in this group to at most n.
A negative value indicates no limit. A limit of zero will prevent any new
goroutines from being added.

Any subsequent call to the Go method will block until it can add an active
goroutine without exceeding the configured limit.

The limit must not be modified while any goroutines in the group are active.
```

### Evidence 7 — canonical godoc spec

Source: `https://go.dev/doc/comment` (Go 1.19+ official guide):

> "A paragraph is a span of unindented non-blank lines."
> "Gofmt replaces duplicated blank lines between paragraphs with a single blank line."

Source: `https://go.dev/blog/godoc` (2011 original):

> "Subsequent lines of text are considered part of the same paragraph; you must leave a blank line to separate paragraphs."

Both sources are **normative**. Any tool that refuses the paragraph break is diverging from the reference definition of Go doc syntax.

---

## Why this rule causes real pain

Captured from the `linter-feedback.md` campaign log (ordered by time sink):

1. Agents reported this as the **single biggest comment-rule time sink** in the 18-package run.
2. Collapsing forces one of three bad outcomes:
   - **Run-on sentence** — fuses 2-3 invariants into one 50+ word sentence (see `WithPublishUpdater` above).
   - **Semicolon chain** — `…A; B; C.` loses the rhetorical "and now a new topic" signal.
   - **Dropped content** — agents drop the secondary paragraph entirely because they can't find a good way to semicolon-in the "See TS 33.128" reference. The code ends up *less* documented after lint.
3. The doc written BEFORE the lint (often by the domain expert who added the function) is almost always more readable than the post-collapse version. The linter degrades documentation quality.
4. It contradicts `KTN-COMMENT-STRUCT` (phase 7) which already tolerates blank-separated paragraphs on types. Two doc-family rules should agree.

---

## Current workaround

Collapse paragraphs via one of:

- Join with semicolons: `…does X; does Y; does Z.`
- Join with explanatory connectives: `…does X and must only be called …`
- Delete the secondary paragraph (worst outcome, applied silently in ~30% of observed cases).

All three lose structural clarity for zero safety benefit.

---

### A. Accept `//` blank-line separators in the description (RECOMMENDED)

Treat `Params:` and `Returns:` as **section headers anchored to line start**, not as "the next non-blank line after the first line of the description". The parser becomes:

```
1. Scan all `//` comment lines preceding `func …`.
2. Split into ordered blocks separated by blank `//` lines.
3. Locate the FIRST block whose first line matches `^Params:\s*$`
   → everything preceding = description (≥ 1 block).
4. Locate the NEXT block whose first line matches `^Returns:\s*$`
   → between it and Params: = param list.
5. Everything after Returns: = return-value list.
```

This change preserves every currently-accepted doc (single-paragraph before `Params:` still works — step 3 finds it immediately) and additionally accepts N-paragraph descriptions (step 3 finds it after N-1 blocks).

Estimated code change: ~20 lines in the parser, one new unit test covering a 2-paragraph and a 5-paragraph description.

### B. Treat `Params:` / `Returns:` as headers, not inline keywords

Anchor them to line start (`regexp: (?m)^// Params:\s*$` and `(?m)^// Returns:\s*$`). This makes the detection independent of whatever comes before, and lets users legitimately mention the word "params" in a description paragraph without triggering false positives.

### C. Match godoc v3 tolerance

Go 1.19 (`https://go.dev/doc/comment`) formalised doc comments as a markdown-ish grammar with explicit paragraph / list / heading rules. Godoc `cmd/gofmt` handles multi-paragraph flawlessly. The linter should match that tolerance so that KTN-COMMENT-FUNC is a **super-set** of godoc validity (adding `Params:` / `Returns:` requirements) rather than a **conflicting** set.

### D. Optional: recognise a "synopsis" vs "body" split (nice-to-have)

`godoc` already treats the **first sentence** as the synopsis (used in package index listings). The linter could enforce a stricter shape for the synopsis (≤ 1 paragraph, ends with a period, starts with the function name) while accepting arbitrary paragraph structure in the body. This matches how stdlib functions are written and how `go doc` renders them.

---

## Back-compat impact

Every doc currently accepted remains accepted — the proposed parser relaxes an adjacency heuristic but does not tighten anything. Concretely:

| Doc shape                                   | Current | Proposed |
|---------------------------------------------|---------|----------|
| 1-paragraph desc + Params + Returns         | ✅      | ✅       |
| 2-paragraph desc + Params + Returns         | ❌      | ✅       |
| 5-paragraph desc + Params + Returns         | ❌      | ✅       |
| No Params section (void function no-arg)    | ✅*     | ✅       |
| Params/Returns with typos ("Parmas:")       | ❌      | ❌       |
| Missing description                         | ❌      | ❌       |

\* — accepted per current rule exemptions for no-arg / no-return functions.

Zero regressions, one large category of false positives removed.

---

## Acceptance criteria

- [ ] `KTN-COMMENT-FUNC` accepts a function comment with N blank-`//` paragraph breaks in the description, provided `Params:` and `Returns:` (when required) still appear as anchored section headers.
- [ ] Unit test: function with 2-paragraph desc → no violation.
- [ ] Unit test: function with 5-paragraph desc (ported verbatim from `os/exec.Cmd.Wait`) → no violation.
- [ ] Unit test: function with typo'd section header ("Parmas:") → still reports violation (regression guard).
- [ ] Unit test: function with description but no `Params:` block when params present → still reports violation (regression guard).
- [ ] `make lint` on `/workspace/services/x2/internal/application/command/enrich_target.go` with `WithPublishUpdater` re-expanded to two paragraphs → passes.

---

## Labels

`enhancement`, `linter-rule`, `false-positive`, `documentation`, `parser`

---

## Related issues

- Conflict #4 (KTN-COMMENT-BLOCK) — sibling over-strict comment rule, forcing trivial `//:` comments.
- Conflict #9 (KTN-COMMENT-STRUCT) — related rule that already tolerates multi-paragraph shape on types; the asymmetry with COMMENT-FUNC is what makes the current behaviour most surprising.

---

## References

- Go Doc Comments (official, Go 1.19+): https://go.dev/doc/comment
- Godoc: documenting Go code (2011 blog post): https://go.dev/blog/godoc
- `context.WithCancel` — https://pkg.go.dev/context#WithCancel
- `context.WithDeadline` — https://pkg.go.dev/context#WithDeadline
- `os/exec.Cmd.Wait` — https://pkg.go.dev/os/exec#Cmd.Wait
- `os/exec.Command` — https://pkg.go.dev/os/exec#Command
- `net/http.Server.Shutdown` — https://pkg.go.dev/net/http#Server.Shutdown
- `net/http.Server.Serve` — https://pkg.go.dev/net/http#Server.Serve
- `golang.org/x/sync/errgroup.Group.Go` — https://pkg.go.dev/golang.org/x/sync/errgroup#Group.Go
- `golang.org/x/sync/errgroup.Group.SetLimit` — https://pkg.go.dev/golang.org/x/sync/errgroup#Group.SetLimit
- In-repo rule catalog: `mcp__ktn-linter__rules code=KTN-COMMENT-FUNC`
- In-repo conflict diagnosis (§Conflict 3): `/workspace/.claude/contexts/ktn-linter-rule-conflicts.md`

### LINKED ISSUE #195: KTN-STRUCT-ROLE: mandatory suffix clashes with Go's `-er` convention and stdlib naming
**Labels**: `enhancement`, `linter-rule`, `false-positive`
**Severity (observed)**: medium — does not block compile, but forces every impl type to grow a redundant suffix, producing stuttering names (`ReloaderService`, `BreakerService`, `CounterAdder`) that do not appear anywhere in the Go standard library.
**Phase (STRUCT-ROLE)**: 1 (warning)
**Phase (INTERFACE-ERNAME)**: 6 (info)

---

## Summary

`KTN-STRUCT-ROLE` (phase 1, warning) requires every exported struct to end in a recognized role suffix (`Service`, `Manager`, `Handler`, `Factory`, `Builder`, `Config`, `Request`, `Response`, `DTO`, `Event`, `Message`, `Result`, etc.).

`KTN-INTERFACE-ERNAME` (phase 6, info) requires single-method interfaces to follow Go's `-er` naming convention (`Reader`, `Writer`, `Stringer`, `Reloader`, `Incer`).

These rules do **not directly contradict**, but when a struct implements an `-er` interface — the canonical Go idiom described in Effective Go and used throughout the standard library — STRUCT-ROLE forces a second suffix on the implementation, producing stuttering names that are **not representative of idiomatic Go**:

| Interface (ERNAME happy) | Impl name STRUCT-ROLE forces | Idiomatic Go name |
|---|---|---|
| `Reloader`   | `ReloaderService`        | `Reloader` or `reloader` |
| `Incer`      | `IncerService`           | `noopCounter` / `counter` |
| `Adder`      | `NoopRevocationCounter`  | `noopAdder` |
| `Breaker`-like iface `Breaker` (Allow/Success/Failure) | `BreakerService` | `Breaker` |

The suffix carries **no additional semantics** — the interface already documents the role. STRUCT-ROLE's suffix list is also **narrower than the stdlib's own naming vocabulary** (`Reader`, `Buffer`, `Client`, `Server`, `Pool`, `Pipe`, `Cond`, `Once`, `Ticker`, `Timer`, `Map`, `Group` are all absent from the list but pervasive in `net/http`, `sync`, `bufio`, `bytes`, `io`, `time`).

**TL;DR** — Either whitelist the `-er` idiom when a matching interface exists in the package, or expand the recognized-suffix list to cover stdlib-common names. The current state forces every codebase that follows Effective Go to add type aliases (`type Reloader = ReloaderService`) as backward-compatibility shims, which is both noise and a tacit admission that the canonical name is the one STRUCT-ROLE rejected.

---

### Repro (minimal)

```go
package litls

// INTERFACE-ERNAME (info, phase 6) is happy: single-method + -er suffix.
type Reloader interface {
    Reload() error
}

// STRUCT-ROLE (warning, phase 1) rejects bare "Reloader":
// "exported struct must end in a recognized role suffix
// (Service, Manager, Handler, Factory, Builder, Config, ...)".
//
// To silence STRUCT-ROLE you MUST rename to ReloaderService:
type ReloaderService struct {
    /* ... */
}

// Now the impl struct name stutters against the interface it implements.
// Effective Go says the struct SHOULD just be named "Reloader" (or unexported
// "reloader") so that callers see `var _ Reloader = (*Reloader)(nil)` — but
// this is impossible under STRUCT-ROLE.

// To keep back-compat with callers who already type `*Reloader`, a type alias
// is inevitably added:
type Reloader = ReloaderService  // ← pure lint-appeasement boilerplate
```

Every `-er` type in the package inherits the same pattern:

```go
// pkg/li/transport/tls/breaker.go:28-40
type BreakerService struct { /* ... */ }
type Breaker = BreakerService  // ← alias for back-compat

// pkg/li/transport/tls/revocation.go:114-127
type RevocationCheckerService struct { /* ... */ }
type RevocationChecker = RevocationCheckerService  // ← alias for back-compat

// pkg/li/transport/tls/reload.go:13-21
type CertWatcherService struct { /* ... */ }
type CertWatcher = CertWatcherService  // ← alias for back-compat
```

### Why the alias is evidence of the clash

Nobody writes `type Foo = FooService` because they like typing. The alias exists **solely because callers upstream already name the type canonically** (`*Reloader`, `*Breaker`) — that is the name Go convention produces, and the linter is the only reason the canonical name is not the primary declaration.

---

## Go idiom evidence

Effective Go, the foundational Go style document, prescribes exactly the naming that STRUCT-ROLE rejects.

### Effective Go — interface names section

Source: <https://go.dev/doc/effective_go#interface-names>

> By convention, one-method interfaces are named by the method name plus an -er suffix or similar modification to construct an agent noun: **Reader**, **Writer**, **Formatter**, **CloseNotifier** etc.
>
> There are a number of such names and it's productive to honor them and the function names they capture. **Read**, **Write**, **Close**, **Flush**, **String** and so on have canonical signatures and meanings. To avoid confusion, don't give your method one of those names unless it has the same signature and meaning. Conversely, **if your type implements a method with the same meaning as a method on a well-known type, give it the same name and signature**; call your string-converter method String not ToString.

The last sentence is the operative one: the implementation is encouraged to mirror the interface's agent noun. Stdlib `bufio.Reader` implements `io.Reader`; the struct is literally named `Reader`. That struct would be flagged by STRUCT-ROLE today.

### Stdlib counterexamples (all would violate STRUCT-ROLE)

| Type | Package | Kind | Matches STRUCT-ROLE suffix list? |
|---|---|---|---|
| `bufio.Reader`        | `bufio`   | struct | no — bare `Reader` |
| `bufio.Writer`        | `bufio`   | struct | no — bare `Writer` |
| `bufio.Scanner`       | `bufio`   | struct | no — bare `Scanner` (agent noun) |
| `bytes.Buffer`        | `bytes`   | struct | no — `Buffer` not in list |
| `bytes.Reader`        | `bytes`   | struct | no — bare `Reader` |
| `io.PipeReader`       | `io`      | struct | no — `PipeReader` |
| `io.PipeWriter`       | `io`      | struct | no — `PipeWriter` |
| `io.SectionReader`    | `io`      | struct | no |
| `io.LimitedReader`    | `io`      | struct | no |
| `http.Server`         | `net/http`| struct | no — `Server` not in list |
| `http.Client`         | `net/http`| struct | no — `Client` not in list |
| `http.Transport`      | `net/http`| struct | no |
| `http.Cookie`         | `net/http`| struct | no |
| `net.Dialer`          | `net`     | struct | no — bare `-er` |
| `net.Listener`        | `net`     | iface+impls | — |
| `net.Conn`            | `net`     | iface+impls (`TCPConn`, `UDPConn`) | no — `Conn` not in list |
| `sync.Mutex`          | `sync`    | struct | no — `Mutex` not in list |
| `sync.RWMutex`        | `sync`    | struct | no |
| `sync.WaitGroup`      | `sync`    | struct | no — `Group` not in list |
| `sync.Once`           | `sync`    | struct | no — `Once` not in list |
| `sync.Cond`           | `sync`    | struct | no — `Cond` not in list |
| `sync.Pool`           | `sync`    | struct | no — `Pool` not in list |
| `sync.Map`            | `sync`    | struct | no |
| `time.Timer`          | `time`    | struct | no |
| `time.Ticker`         | `time`    | struct | no |
| `time.Location`       | `time`    | struct | no |
| `strings.Builder`     | `strings` | struct | yes (`Builder`) |
| `strings.Reader`      | `strings` | struct | no |
| `strings.Replacer`    | `strings` | struct | no — agent noun `-er` |
| `text/template.Template` | `text/template` | struct | no |
| `html/template.Template` | `html/template` | struct | no |
| `log.Logger`          | `log`     | struct | no — bare `-er` |
| `log/slog.Logger`     | `log/slog`| struct | no — bare `-er` |
| `log/slog.Handler`    | `log/slog`| interface | — |
| `crypto/tls.Conn`     | `crypto/tls`| struct | no |
| `crypto/tls.Config`   | `crypto/tls`| struct | yes (`Config`) |
| `crypto/tls.CertificateRequestInfo` | `crypto/tls` | struct | no |
| `crypto/x509.Certificate` | `crypto/x509` | struct | no |
| `crypto/x509.CertPool`    | `crypto/x509` | struct | no |
| `encoding/json.Decoder` | `encoding/json` | struct | no — agent noun |
| `encoding/json.Encoder` | `encoding/json` | struct | no — agent noun |
| `context.Context`     | `context` | interface | — |
| `errors.*` (no structs) | | | |

Approximate tally: **>80% of exported structs in the main stdlib packages violate STRUCT-ROLE's current suffix list**. The rule, applied to stdlib source, would flag the vast majority of its public surface.

Reference URLs:
- <https://pkg.go.dev/bufio#Reader> — confirm struct is bare `Reader`
- <https://pkg.go.dev/bytes#Buffer> — confirm struct is bare `Buffer`
- <https://pkg.go.dev/net/http#Server> — confirm struct is bare `Server`
- <https://pkg.go.dev/net/http#Client> — confirm struct is bare `Client`
- <https://pkg.go.dev/sync#Mutex> — confirm bare `Mutex`
- <https://pkg.go.dev/sync#WaitGroup> — confirm bare `WaitGroup`
- <https://pkg.go.dev/time#Timer> — confirm bare `Timer`
- <https://pkg.go.dev/log/slog#Logger> — confirm bare `Logger`
- <https://go.dev/doc/effective_go#interface-names> — the operative quote
- <https://go-proverbs.github.io> — Rob Pike: "Don't name a type for what it is, name it for what it does." (implicitly forbids `Service` stuffing).

### Quote — Go Code Review Comments (interface names)

Source: <https://go.dev/wiki/CodeReviewComments#interface-names>

> By convention, one-method interfaces are named by the method name plus an -er suffix or similar modification to construct an agent noun: Reader, Writer, Formatter, CloseNotifier etc. There are a number of such names and it's productive to honour them and the function names they capture. Read, Write, Close, Flush, String and so on have canonical signatures and meanings.

This is reiterated in <https://go.dev/blog/package-names> ("Avoid meaningless package names like util, common, misc"). The same logic applies to type names — `FooService` is the type-level analogue of a `foo_util` package: it adds a non-semantic decorator instead of letting the interface describe the role.

---

### Example 1 — `pkg/li/transport/tls/reloader.go`

`/workspace/pkg/li/transport/tls/reloader.go`

```go
// line 20 — single-method interface, ERNAME-compliant
type Incer interface {
    Inc()
}

// line 92 — STRUCT-ROLE-forced name
type ReloaderService struct {
    certPath       string
    keyPath        string
    log            *slog.Logger
    pollInterval   time.Duration
    successCounter Incer
    failedCounter  Incer
    // ...
}

// line 108 — the back-compat alias that betrays the clash
// Reloader is a backward-compatible alias for ReloaderService.
type Reloader = ReloaderService
```

The package also exports `NewReloaderService` (line 120) **and** `NewReloader` (line 172) as a forwarding constructor. Two public constructors for one type — one named per convention, one named per the linter.

`noopCounter` (file `/workspace/pkg/li/transport/tls/reloader_noop.go`) is unexported and therefore not flagged by STRUCT-ROLE. Had it been exported, it would have become `NoopCounterService` — stuttering twice over (`Noop`+`Counter`+`Service`) for a 1-line type with one 0-statement method.

### Example 2 — `pkg/li/transport/tls/breaker.go`

```go
// line 28 — STRUCT-ROLE-forced name
type BreakerService struct { /* circuit-breaker state machine */ }

// line 40 — back-compat alias
// Breaker is a backward-compatible alias for BreakerService.
type Breaker = BreakerService
```

`Breaker` is a **stdlib-style name** (cf. `time.Timer`, `time.Ticker`, `sync.Pool`). The alias exists because every callsite in `pkg/li/transport/tls/client.go` uses `*Breaker` (the idiomatic name), not `*BreakerService` — see lines 73 and 162 of `/workspace/pkg/li/transport/tls/client.go`.

### Example 3 — `pkg/li/transport/tls/revocation.go`

```go
// line 65 — single-method interface (the -er is "WithLabelValueser",
//   which is itself a mark of strain: applying the -er rule mechanically
//   to a method called WithLabelValues produces an ugly name).
type WithLabelValueser interface {
    WithLabelValues(lvs ...string) Adder
}

// line 71 — another canonical -er interface
type Adder interface {
    Add(float64)
}

// line 114 — STRUCT-ROLE-forced name
type RevocationCheckerService struct { /* ... */ }

// line 127 — back-compat alias
type RevocationChecker = RevocationCheckerService
```

`RevocationChecker` is already an `-er` form. STRUCT-ROLE still forces `RevocationCheckerService` — an `-er`-suffix noun followed by a role suffix, which is the stuttering pattern Effective Go warns against.

### Example 4 — `pkg/li/transport/tls/reload.go`

```go
// line 13
type CertWatcherService struct { /* fsnotify-backed watcher */ }

// line 21
type CertWatcher = CertWatcherService
```

Again: `-er` noun + `Service` stuffing + back-compat alias.

### Example 5 — `test/e2e/mocks/df2/server/store.go`

`/workspace/test/e2e/mocks/df2/server/store.go`

```go
// line 20
type StoreService struct {
    mu    sync.RWMutex
    pdus  []ReceivedPDUMessage
    stats StatsSpec
}

// line 30
func NewStoreService() (store *StoreService) { /* ... */ }
```

No interface named `Storer` exists; `StoreService` was chosen purely to satisfy STRUCT-ROLE. Compare with stdlib `sync.Map`, `bytes.Buffer`, `x509.CertPool` — all bare nouns, none with a `Service` suffix. The idiomatic name would be `Store` (cf. <https://pkg.go.dev/context#Context> which uses bare nouns throughout).

### Example 6 — `test/e2e/mocks/df2/server/certrotator.go`

```go
// line 21
type CertRotatorService struct {
    current atomic.Pointer[CertBundleSpec]
}
```

`CertRotator` would be an `-er`-form agent noun (one who rotates certs). STRUCT-ROLE forces `CertRotatorService`. The type has three methods (`Load`, `Swap`, `Certificate`) — not a single-method interface; but the NAME is still an `-er` agent noun for which STRUCT-ROLE provides no whitelist.

### Example 7 — `test/e2e/mocks/df2/server/listener.go`

`ListenerService` (line 24). The stdlib's `net.Listener` is an interface; concrete implementations are `*net.TCPListener`, `*net.UnixListener`. STRUCT-ROLE, applied to stdlib, would rename these to `TCPListenerService` / `UnixListenerService`.

---

## Proposed fix

Three alternatives, ordered by ambition. (A) and (B) can be combined; (C) is an orthogonal severity change.

### A. Whitelist the `-er` stem when the package declares a matching interface

**Mechanic**: if a struct `T` is named with an `-er` suffix (`Reloader`, `Breaker`, `Rotator`, `Watcher`, `Checker`, `Reader`, `Writer`, etc.) AND the package (or an imported package referenced in a compile-time assertion `var _ I = (*T)(nil)`) declares an interface with the **same stem** (`Reloader interface`, `Breaker interface`, …), accept the struct name without a role suffix.

**Detection heuristics** (rank by certainty):

1. `var _ <IfaceName> = (*<T>)(nil)` or `var _ <IfaceName> = <T>{}` appears in the same package.
2. The package declares `type <IfaceName> interface { … }` AND `T` has a method set that covers `<IfaceName>`'s method set (requires type-check info; cheap post-`go/types` resolution).
3. The struct name ends in one of the canonical `-er` stems (`Reader`, `Writer`, `Closer`, `Stringer`, `Formatter`, `Flusher`, `Handler` — already accepted —, `Watcher`, `Rotator`, `Checker`, `Breaker`, `Builder` — already accepted —, `Scanner`, `Parser`, `Encoder`, `Decoder`, `Marshaler`, `Unmarshaler`, `Dialer`, `Listener`, `Logger`).

Rule (1) alone covers the strongest idiomatic signal (explicit compile-time assertion) with zero false positives.

**Benefit**: the back-compat aliases (`type Reloader = ReloaderService`, `type Breaker = BreakerService`, etc.) disappear. The canonical name becomes the primary declaration, matching stdlib practice.

**Effort**: small. The linter already walks type declarations; adding a pass that indexes interface names per package and cross-checks struct names is a standard AST visit.

### B. Expand the recognized-suffix list to include stdlib-common names

Add to STRUCT-ROLE's accepted suffix set:

- **Containers/primitives**: `Buffer`, `Pool`, `Cache`, `Map`, `List`, `Queue`, `Stack`, `Set`, `Ring`, `Heap`, `Tree`
- **Concurrency**: `Mutex`, `RWMutex`, `Once`, `Group`, `WaitGroup`, `Cond`, `Semaphore`, `Limiter`, `Barrier`
- **Network/IO**: `Client`, `Server`, `Conn`, `Connection`, `Dialer`, `Listener`, `Transport`, `Session`, `Reader`, `Writer`, `Scanner`, `Parser`, `Encoder`, `Decoder`, `Pipe`
- **Time**: `Timer`, `Ticker`, `Clock`, `Deadline`
- **Crypto/Auth**: `Cert`, `CertPool`, `Certificate`, `Signer`, `Verifier`, `Key`, `Keystore`
- **Generic `-er` agent nouns** (regex `.+er$`): the family Effective Go endorses

The key insight: STRUCT-ROLE's suffix list currently **encodes J2EE-style DDD vocabulary** (`Service`, `Manager`, `Factory`, `Builder`, `Handler`, `Config`, `DTO`, `Request`, `Response`, `Event`, `Message`, `Error`, `Result`) — an inheritance from enterprise-Java pattern catalogs — and misses most of the Go-idiomatic vocabulary. Expanding the list to cover stdlib names brings the rule in line with the language it lints.

### C. Downgrade STRUCT-ROLE to `info` when interface stem matches

If proposals (A) and (B) are too invasive for a single change, a minimal fix: when struct `T`'s name matches the stem of an interface in the same package (`T=Reloader` and `interface Reloader` both exist), downgrade STRUCT-ROLE from `warning` to `info` with the message `"idiomatic -er impl of local interface; consider keeping the bare name"`. Pre-commit hooks that block on warnings but allow info-level then stop being the deciding factor; teams can opt into the idiom.

---

## Why this matters

1. **Linter output should match Go conventions the compiler and community already codify.** A rule that flags 80%+ of stdlib names is miscalibrated against its target language.
2. **Back-compat aliases are lint-induced tech debt.** Every `type Reloader = ReloaderService` in this codebase exists because callers wrote the idiomatic name first, the linter objected, and the author punted to a transparent alias. The alias will live forever (removing it breaks callers); the `Service` suffix will live forever (STRUCT-ROLE demands it). The package ends up with two names for every type.
3. **ERNAME and STRUCT-ROLE were designed independently.** ERNAME (phase 6, info) endorses the `-er` idiom. STRUCT-ROLE (phase 1, warning) ignores it. The two rules should at minimum consult each other: when ERNAME would be happy with `TypeName`, STRUCT-ROLE should not contradict it.
4. **Readability regression.** `ReloaderService.Reload()`, `BreakerService.Allow()`, `RevocationCheckerService.VerifyPeerCertificate()` — these are all longer than the stdlib equivalents (`Reader.Read`, `Server.ListenAndServe`, `Logger.Info`) without adding information. The `Service` decorator is signal-free suffix noise.

---

## Evidence summary table

| File | Line | Canonical name (Go idiom) | STRUCT-ROLE-forced name | Has `-er` iface in pkg? | Has alias? |
|---|---|---|---|---|---|
| `pkg/li/transport/tls/reloader.go`  | 92  | `Reloader`           | `ReloaderService`         | `Incer` (related) | yes (L108) |
| `pkg/li/transport/tls/breaker.go`   | 28  | `Breaker`            | `BreakerService`          | — (concrete type) | yes (L40)  |
| `pkg/li/transport/tls/revocation.go`| 114 | `RevocationChecker`  | `RevocationCheckerService`| `Adder`, `WithLabelValueser` | yes (L127) |
| `pkg/li/transport/tls/reload.go`    | 13  | `CertWatcher`        | `CertWatcherService`      | — | yes (L21)  |
| `test/e2e/mocks/df2/server/store.go`| 20  | `Store`              | `StoreService`            | — | no         |
| `test/e2e/mocks/df2/server/certrotator.go` | 21 | `CertRotator` | `CertRotatorService`      | — | no         |
| `test/e2e/mocks/df2/server/listener.go` | 24 | `Listener` (cf. `net.Listener`) | `ListenerService` | — | no |

Four of seven examples carry a type alias **solely to work around STRUCT-ROLE**. The other three lack an alias only because they are test-only mocks with fewer external callers.

---

## Acceptance criteria for fix

- [ ] Given `type Reloader interface { Reload() error }` and `type Reloader struct { … }` in the same package, STRUCT-ROLE does **not** fire (proposal A).
- [ ] Given `type Server struct { … }` (no matching local interface), STRUCT-ROLE does **not** fire (proposal B — `Server` added to suffix list).
- [ ] Given `type Adder interface { Add(float64) }` and `type NoopAdder struct{}` with `func (NoopAdder) Add(float64) {}`, STRUCT-ROLE does **not** fire (stem match via method-set resolution — proposal A).
- [ ] Backwards compatibility: `FooService`, `FooManager`, etc. remain valid (nobody needs to rename).
- [ ] Existing test fixtures covering STRUCT-ROLE are updated to reflect that `Reader`, `Server`, `Client`, `Buffer`, `Timer` are now accepted bare (proposal B).
- [ ] ERNAME-compatible names (`-er` suffix) are whitelisted when a matching interface exists in the package (proposal A).
- [ ] The 4 back-compat aliases in `/workspace/pkg/li/transport/tls/*.go` can be removed without adding STRUCT-ROLE warnings.

---

## Related rule-conflict entries

- Context catalog: `/workspace/.claude/contexts/ktn-linter-rule-conflicts.md` § Conflict 5
- Conflict #1 (Unused-arg quadrilemma) — also exacerbated by `-er` impls: methods forced to accept interface-required args they ignore (e.g. `ReloaderService.Certificate(*tls.ClientHelloInfo)`).
- Conflict #6 (NAMEDPARAMS on func types) — adjacent: same enterprise-Java vs idiomatic-Go axis.

---

## References

- Effective Go — Interface names: <https://go.dev/doc/effective_go#interface-names>
- Go Code Review Comments — Interface names: <https://go.dev/wiki/CodeReviewComments#interface-names>
- Go package names blog post: <https://go.dev/blog/package-names>
- Go Proverbs: <https://go-proverbs.github.io>
- `bufio.Reader`: <https://pkg.go.dev/bufio#Reader>
- `bytes.Buffer`: <https://pkg.go.dev/bytes#Buffer>
- `net/http.Server`: <https://pkg.go.dev/net/http#Server>
- `net/http.Client`: <https://pkg.go.dev/net/http#Client>
- `sync.Mutex`: <https://pkg.go.dev/sync#Mutex>
- `sync.WaitGroup`: <https://pkg.go.dev/sync#WaitGroup>
- `time.Timer`: <https://pkg.go.dev/time#Timer>
- `log/slog.Logger`: <https://pkg.go.dev/log/slog#Logger>
- `net.Dialer`: <https://pkg.go.dev/net#Dialer>
- `crypto/tls.Config`: <https://pkg.go.dev/crypto/tls#Config>
- `encoding/json.Decoder`: <https://pkg.go.dev/encoding/json#Decoder>

kodflow:
## Summary

Executes the P0 plan in `.claude/plans/p0-linter-bugs.md`. 4 P0 bugs were investigated; 3 land here, 1 (`#183`) deferred to a dedicated PR.

| # | Title | Status | Commit |
|---|---|---|---|
| **#193** | KTN-COMMENT-FUNC parser multi-paragraph | **Closed** (already fixed in `74f20787`; this PR adds stdlib regression fixtures) | `c665df99` |
| **#182** | KTN-TEST-DECISION hardcoded modulePrefix | **Closed** (already fixed in `b380bf02`; this PR adds foreign-module regression tests for `walkToGoMod`) | `672ad464` |
| **#195** | KTN-STRUCT-ROLE vs `-er` convention | **Closed** (real fix: TierStdlib whitelist + interface-impl exemption) | `f288743f` |
| #183 | KTN-VAR-MAKEAPPEND/SLICECAP false positives | **Deferred** | Needs Object-identity refactor that breaks ~50 internal-test sites that build `map[string]bool{...}` literals directly. Cleanest in its own focused PR. |

Closes #182. Closes #193. Closes #195.

## #195 — STRUCT-ROLE fix (the real change)

Two complementary mechanisms (Option C from the plan):

### A — auto-exempt structs implementing a sibling interface
- `pkg/analyzer/ktn/structure/organization/role.go::implementsLocalInterface`
- Walks `pass.Pkg.Scope()`, calls `go/types.Implements` against every non-empty interface declared in the same package.
- Honours pointer-receiver method sets via `types.NewPointer(named)`.
- Skips uninstantiated generics where `Implements` is unspecified.

### B — TierStdlib whitelist (24 stdlib-canonical suffixes)
- `pkg/analyzer/shared/roles.go`: new `TierStdlib` constant + 24 entries (`Reader`, `Writer`, `Scanner`, `Encoder`, `Decoder`, `Logger`, `Dialer`, `Listener`, `Closer`, `Buffer`, `Pool`, `Cache`, `Pipe`, `Conn`, `Client`, `Server`, `Group`, `Once`, `Cond`, `Mutex`, `RWMutex`, `Ticker`, `Timer`, `Map`).
- `Suggestible: false` so `InferRole` never proposes them — they exist purely to ACCEPT the names without forcing a Service/Handler suffix.
- `reportDiscouraged` already only fires for `TierDiscouraged` so TierStdlib is silent-accept by construction.

## Test plan

- [x] `go build ./...` — green
- [x] `go test ./...` — all packages green (exits 0 across all 60+ packages)
- [x] `pkg/analyzer/shared/roles_{internal,external}_test.go` — `expectedRoleCount` 24 → 48
- [x] `pkg/analyzer/ktn/structure/organization/testdata/src/structrole/good.go` — new fixtures: `Notifier`+`Pinger` (Option A), `Buffer`, `Server`, `Reader`, `Pool` (Option B). `BadUser` still flagged.
- [x] `pkg/analyzer/ktn/comment/docs/testdata/src/commentfunc/{good,bad}.go` — 3 stdlib godoc regression fixtures + 1 negative case.
- [x] `pkg/analyzer/ktn/tests/quality/decision_internal_test.go` — `Test_walkToGoMod_ForeignModule` (4 sub-tests covering `example.com/other`, `gitlab.example.com/team/project/service`, replace-style, ktn-linter regression) + `Test_walkToGoMod_NoGoMod` fail-open.

## #183 deferral rationale

Issue #183 (KTN-VAR-MAKEAPPEND/SLICECAP false positives) requires migrating `appendVars map[string]bool` to `map[types.Object]bool` to defeat cross-function name collisions. The migration touches ~25 production sites + ~50 internal-test sites that construct map literals directly (`map[string]bool{"x": true}`). A simpler "augment the collector" fix is unsafe because the gate logic is `!appendVars[name] → skip`: adding more entries makes MORE diagnostics fire, not fewer. Cleanest path is a dedicated PR with the full Object refactor + test updates.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

- KTN-COMMENT-FUNC (KTN-COMMENT-FUNC, #193): Parser relaxed to accept multi-paragraph // doc comments (split into blocks by blank // lines); anchored Params:/Returns: headers still recognized and description is the blocks before the first Params: block. Added good/bad regression fixtures (multi-paragraph success cases and a stdlib-shaped misname negative case) and updated analyzer expectations.

- KTN-STRUCT-ROLE (KTN-STRUCT-ROLE, #195): Two-part fix — (1) implementsLocalInterface uses go/types.Implements (honors pointer receivers, skips uninstantiated generics) to auto-exempt exported structs that satisfy a non-empty sibling interface; (2) introduced TierStdlib with 24 stdlib-canonical suffixes (e.g., Reader, Writer, Buffer, Pool, Mutex, Client, Server, Conn, Listener, Timer, Ticker, Map, Group) registered as non-suggestible so common stdlib names no longer trigger role-suffix diagnostics. Tests/fixtures added; shared.RoleCount updated accordingly.

- KTN-TEST-DECISION (KTN-TEST-DECISION, #182): Removed hardcoded modulePrefix by implementing walkToGoMod to locate and parse go.mod to derive module path and root; threaded resolved module root into runtime-coverage helpers. Added table-driven tests for foreign-module prefixes and strict "no go.mod" fail-open semantics; test updated to enforce strict behavior (t.Fatalf).

- Tests, messages, and registry impact: Expanded role registry (shared.RoleCount → 48) and updated role-related tests; broad regression test additions/updates across commentfunc, structrole, and walkToGoMod suites. Full build and go test across packages passed locally.

- No govet/modernize wrapper, CLI behavior, orchestrator, or cache changes introduced in this PR; KTN-VAR-MAKEAPPEND/SLICECAP fix (#183) was deferred for a follow-up PR.


kodflow:
## CodeRabbit triage round 1 — `56e0dceb`

5 actionable findings on `f288743f`. 3 LEGITIMATE fixed, 2 rejected with justification.

### Fixed

| File:Loc | Finding | Fix |
|---|---|---|
| `func_external_test.go:32-38` | Subtest needs `t.Parallel()` first | Added as first statement in the `t.Run` closure body. ✓ |
| `roles.go:102-103` | Doc comment "= 47" is stale | Corrected to "= 48" (24 historical + 24 stdlib). ✓ |
| `decision_internal_test.go::Test_walkToGoMod_NoGoMod` | Test was loose; should assert strict `gotPrefix == "" && gotRoot == ""` | Rewritten table-driven with strict fail-open assertion + `t.Skipf` fallback for CI runners that have a go.mod above `os.TempDir`. ✓ |

### Rejected (style preferences)

| File:Loc | Finding | Why rejected |
|---|---|---|
| `role.go:166-208` | Refactor `implementsLocalInterface` into smaller helpers (Codacy cyclomatic-complexity warning) | Same Codacy CC nag rejected on PR #216. The project's own `KTN-FUNC-CYCLO` does not flag the current shape (CC=9 within tolerance). The function is one logical operation; splitting it adds indirection without behavioural value. |
| `role.go:135-142` | Narrow auto-exempt to `-er` suffix only | Contradicts the plan's design intent (Option A in `.claude/plans/p0-linter-bugs.md`). The exemption rationale is "the role IS the interface contract" — restricting to `-er` re-introduces the FP for legitimately-named non-er implementations (e.g. `type Foo struct{}` implementing `Notifier`). The plan deliberately chose semantic exemption (interface impl) over syntactic restriction (suffix shape). |

CI re-trigger expected on push.

@coderabbitai review

kodflow:
### Triage round 2 (commit `4c1898b5`)

**1 LEGITIMATE finding fixed:**

- ✅ `decision_internal_test.go` — replaced `t.Skipf` with `t.Fatalf` for strict fail-open contract (CodeRabbit). The fallback skip masked failures when the test environment has a go.mod above `os.TempDir`; it also violated our own KTN-TEST-NOSKIP rule. Now any breach of the `("","")` contract surfaces loudly.

@coderabbitai review

kodflow (commented):
### Review state — supersede stale CHANGES_REQUESTED

CodeRabbit findings across rounds 1–2 have been triaged as follows on the latest commit `4c1898b5`:

**LEGITIMATE — fixed (4):**
- ✅ `func_external_test.go` — `t.Parallel()` placed first inside subtest
- ✅ `roles.go` — doc comment fixed `47` → `48`
- ✅ `decision_internal_test.go` round 1 — table-driven `Test_walkToGoMod_NoGoMod` with strict `("","")` assertion
- ✅ `decision_internal_test.go` round 2 — `t.Skipf` → `t.Fatalf` (also enforces our own KTN-TEST-NOSKIP)

**ILLEGITIMATE — rejected with justification (2):**
- ❌ `role.go::implementsLocalInterface` cyclomatic-complexity refactor — same Codacy CC nag pattern that was rejected on PR #216 with the documented rationale: extraction would obscure the type-checker traversal, the function is read end-to-end as a unit, and CC alone is not a quality signal here.
- ❌ Restrict the local-interface exemption to `-er` names only — directly contradicts plan AC-7 ("Buffer/Pool/Server/Reader/Pinger that implement a sibling interface must NOT trigger STRUCT-ROLE"). Names like `Buffer`, `Pool`, `Pipe`, `Conn` deliberately exist without the `-er` suffix; they are the very names the stdlib codifies and the plan whitelists. Adding the restriction would re-introduce the false positives this PR is meant to close.

CodeRabbit's review on `4c1898b5` returned "Review completed" with **no new actionable comments**. This COMMENT review supersedes the prior CHANGES_REQUESTED state.

kodflow:
@coderabbitai resolve

All actionable findings on this PR have been triaged and either fixed or rejected with documented justification on commit `4c1898b5`. CodeRabbit's last review check returned "Review completed" success without posting any new actionable comments. Please resolve the prior threads so this PR can merge.

@coderabbitai review

PR #219, opened 2026-05-01
perf(serve): bound scan concurrency + /statusline endpoint + debug tracing

kodflow:
## Why

When the MCP server runs `scan ./...` against this project (`cache: false`, ~70 packages, 403 analyzers), every CPU core gets pinned to 100 % until the user kills the process. The user reported having to kill the daemon. Three orthogonal mitigations are applied here, plus a `/statusline` endpoint for the Claude Code statusline binary the user is building.

Diagnostic written up in [`.claude/contexts/ktn-linter-perf-cache-logs.md`](../blob/perf/scan-bound-concurrency-tracing-statusline/.claude/contexts/ktn-linter-perf-cache-logs.md) — root cause, citations, and ordered plan.

### 1. Perf — bound the analysis fan-out

`pkg/orchestrator/runner.go:70` previously did:

```go
grp.SetLimit(runtime.GOMAXPROCS(0))
```

On a 16-core machine this saturates every core during a fresh full scan with no headroom for the host (IDE, tests, other agents). Replaced with:

```go
grp.SetLimit(effectiveMaxWorkers(r.maxWorkers))   // default: max(2, GOMAXPROCS/2)
```

- `Options.MaxWorkers int` (yaml `performance.max_workers`)
- `AnalysisRunnerService.SetMaxWorkers(int)` setter
- Wired from `OrchestratorService.Run`, `RunMultiModule`, `runSingleModule` on **every** invocation so YAML hot-reload takes effect immediately.

Heuristic source: `~/.claude/docs/concurrency/thread-pool.md` § "When to Use" — CPU-bound work ≤ NumCPU, reserve headroom for the host.

### 2. Perf — fewer parasitic fsnotify events

- Watcher debounce **100 ms → 300 ms** (`pkg/serve/watcher.go:23`). Matches editor-save coalescing in `~/.claude/docs/performance/debounce-throttle.md`.
- `skipDirs` extended (`pkg/serve/watcher.go:31-50`):

  | Added | Rationale |
  |-------|-----------|
  | `bazel-out`, `bazel-bin`, `bazel-testlogs`, `bazel-ktn-linter` | Bazel sandbox + generated sources; every `bazel test` cycle would otherwise fire fsnotify events |
  | `.git` | Noisy under `git rebase`; zero project source |

### 3. `/statusline` endpoint

For `https://github.com/kodflow/status-line` polling. `GET /statusline` returns a compact JSON shape designed for ~100 ms refresh:

```json
{
  "scanning": false,
[...]

PR #220, opened 2026-05-01
perf(serve): scan gate semaphore — bound cross-scan concurrency

kodflow:
## Why

PR #219 capped intra-scan fan-out via `Options.MaxWorkers`, but didn't address the **cross-scan** case: two scans of distinct packages each spin up their own `errgroup.SetLimit(N)` pool, doubling CPU saturation.

Item 7 of `.claude/contexts/ktn-linter-perf-cache-logs.md` §4.1.B — "Mutex global de scan dans `EngineService`".

> "Empêche un `scan ./...` MCP + un fsnotify simultanés de doubler le travail."

## What

Add `EngineService.scanGate chan struct{}` (cap 1) acquired/released inside `beginScan()`:

```go
func (e *EngineService) beginScan() func() {
    if e.scanGate != nil {
        e.scanGate <- struct{}{}    // blocking acquire
    }
    e.scanInFlight.Add(1)
    return func() {
        if e.scanGate != nil {
            <-e.scanGate              // release first, unblocks next waiter
        }
        e.scanInFlight.Add(-1)
    }
}
```

### Design

| Aspect | Choice | Why |
|--------|--------|-----|
| Capacity | 1 | At most one scan at a time across the whole engine |
| Acquisition | Blocking send | Concurrent scans queue up rather than failing or doubling fan-out |
| Order | Release-before-decrement | Next waiter can acquire while we decrement; counter still reaches 0 |
| Nil handling | No-op when `scanGate == nil` | Zero-value `&EngineService{}` (legacy fixtures) still works |
| singleflight interaction | Gate runs INSIDE `sfGroup.Do` callback | Same-key duplicates coalesce *before* the gate so legitimate dedup is preserved |

### Reentrancy split — `beginScan` vs `beginScanCount`

`WarmUp` calls `ScanAll` on the slow path. If both invoked `beginScan`, the second send on the cap-1 channel would deadlock (same goroutine sender + receiver). Split into:

- `beginScan()` — gate + counter, for top-level entries (`scanFresh`, `ScanAll`)
- `beginScanCount()` — counter only, for nested callers (`WarmUp`)

`WarmUp`'s fast paths (cache:false, bolt-fast) don't run analyzers so they don't need the gate; the slow path delegates to `ScanAll` which acquires it.

## Verification

- `make build` → bazel ok
- `make test` → 96/96 bazel tests pass
- New tests:
  - `TestEngineService_ScanGate_Serializes` — two goroutines, second must acquire AFTER first releases (timestamps + controlled holds 50ms, 100ms)
  - `TestEngineService_ScanGate_NilSafe` — zero-value engine still works

## Out of scope

Items 8 (engine cache wiring), 9 (Facts activation), 10 (L2 eviction) tracked separately. This PR only delivers item 7.

Builds on top of #219 (which is awaiting merge).

• **Scan gate semaphore for cross-scan concurrency control**: Introduces a capacity-1 channel (`scanGate`) within EngineService to serialize project-wide scans and prevent CPU overcommit when multiple concurrent scan sources (MCP, HTTP, file watcher) run in parallel. The gate complements the orchestrator's existing intra-scan errgroup fan-out bounds by adding the missing cross-scan concurrency limit that was previously missing.

• **Observability enhancements**: Adds exported methods `IsScanning()` and `IsWarmupDone()` to expose `scanInFlight` counter and warmup completion state; `Status()` now reflects live scanning state; default engine state transitions from `StateWatching` to new `StateInit` constant to distinguish "never scanned" from "finished scanning."

• **Test coverage delta (+4 tests)**: New unit tests cover nil-safety of gate on zero-value EngineService (`TestEngineService_ScanGate_NilSafe`), serialization enforcement for concurrent scans (`TestEngineService_ScanGate_Serializes`), scanning counter observability (`TestEngineService_IsScanning_Counter`), and Status reflection of scanning state (`TestEngineService_Status_ReflectsScanning`); existing state transition tests updated to expect `StateInit` as new default.

• **No KTN rule, severity registry, govet/modernize, or CLI changes**: PR focuses exclusively on EngineService performance optimization and does not introduce new KTN rule definitions, severity registry updates, govet wrapper modifications, or CLI behavior changes.


kodflow (commented):
Bot threads triaged. Legitimate test-flake fix applied (8ea13fa1: `Before` instead of `After`). Other findings are scope-of-#219 territory or already-justified DTO-tag rejections. Superseding the CHANGES_REQUESTED state.

PR #266, opened 2026-05-07
feat(structrole): constructor-as-role-marker exemption (issue #263 phase 3)

kodflow:
## Summary

Phase 3 of the issue #263 redesign. Stacked on top of #265.

Adds the third decision step to `KTN-STRUCT-ROLE`: accept structs whose package-level constructor name expresses the role even when the type name lacks a recognised suffix. This is Effective Go's `bytes.NewBuffer` / `bufio.NewReader` / `regexp.Compile` / `net.Dial` / `os.Open` idiom in package form.

## Decision order in the analyzer

```
1. Recognised role suffix       → accept
2. PNPT exemption               → accept (PR #265)
3. Constructor exemption        → accept (THIS PR)
4. Interface carrier            → accept
5. otherwise                    → emit KTN-STRUCT-ROLE
```

## What's accepted

A struct `T` is exempt iff BOTH signals hold:

1. The package declares a top-level factory whose first return is `T` or `*T` (optionally with `error` second). Recognised names: `New`, `New<TypeName>`, `Open`, `Dial`, `Make`, `Compile`.
2. The struct is **behaviour-bearing**: at least one method (value or pointer receiver) on `T`.

Pure data structs (no methods) deliberately fall through to `KTN-STRUCT-CTOR` / `-JSONTAG` / `-PRIVTAG`. DTO-classified structs (`>= LikelyDTO`) are filtered upstream so JSON-tagged shapes remain governed by `JSONTAG` / `PRIVTAG`.

Generic constructors and tuples where `T` is not the primary return are rejected — narrow signature contract by design.

## Test impact

`structrolector_gates` flips `bad.go` → `good.go`; the test entry's `expected` drops from 1 to 0. The Watcher fixture (in package `gates`, with `NewWatcher` and an `Observe()` method) is the canonical case.

The other corpus buckets remain unchanged: tier guard silent, negative bucket (util/common/helper) firing, DTO guard firing.

## Out of scope

- Generic-parameterised constructor signatures (deferred).
- Wrapper return types like `Result[T]` or `(*Other, *T, error)` (deferred).

## Test plan

- [x] `make build` green.
- [x] `make test` 97/97.
- [x] `TestStructRole` 12 entries — gates flips to expected 0.
- [ ] CI verifies the same on the PR branch.

References: issue #263, plan `.claude/plans/issue-263-struct-role-redesign.md`.

* Adds constructor-as-role-marker exemption to KTN-STRUCT-ROLE (phase 3 of issue #263): exported structs are exempt when (1) package defines a recognized top-level factory (New, New<TypeName>, Open, Dial, Make, Compile) returning T or *T (optionally with error) and (2) the struct has at least one method. Decision order updated: suffix → PNPT exemption → constructor exemption → interface carrier → emit KTN-STRUCT-ROLE.
* Implementation changes: new internal analyzer file constructor.go with hasConstructorRole and signature guards (uses isPNPTConstructorSignature); excludes test files and rejects generic constructors (explicit TypeParams guard) and tuple/non-primary returns by design; pure data structs remain governed by KTN-STRUCT-CTOR / KTN-STRUCT-JSONTAG / KTN-STRUCT-PRIVTAG.
* Tests and fixtures: structrolector_gates case flipped (Watcher moved from bad.go → good.go) so expected KTN-STRUCT-ROLE drops from 1→0; new structrolector_generic testdata added to ensure generic constructors (e.g., NewStash[T]) continue to trigger KTN-STRUCT-ROLE.
* Build/test status and scope: make build OK; make test 97/97 reported locally; CI pending. Out of scope: generic-parameterised constructors and auxiliary/wrapper return types.


kodflow:
@coderabbitai resolved.

Pushed `7cfbca44`:
- `funcDecl.Type.TypeParams != nil` early-skip in `packageHasConstructorReturning` (constructor.go:88).
- New `structrolector_generic` testdata bucket: `Stash[T any]` with `NewStash[T any]()` + `Observe` method must keep firing `KTN-STRUCT-ROLE` (regression guard).
- `make build` + `make test` 97/97 green locally.

PR #268, opened 2026-05-07
feat(structrole): KTN-STRUCT-ROLE-COMPATALIAS companion (issue #263 phase 5)

kodflow:
## Summary

Phase 5 of the issue #263 redesign. Stacked on top of #267.

Adds the **Goodhart-canary self-detecting analyzer**: flags an alias `type X = Y` where Y is a same-package suffixed struct AND the alias's doc comment self-discloses the backward-compat motive. Catches the exact anti-pattern issue #263 §H1 documented — when authors satisfy the suffix taxonomy by renaming `Foo` → `FooService` and adding `type Foo = FooService`, doubling the public surface for zero clarity gain.

## Four-gate predicate

The diagnostic fires iff ALL hold:

1. The alias name is exported AND has no recognised role suffix.
2. The alias target resolves to a same-package suffixed exported struct (cross-package selectors out of scope this phase).
3. The declaration is the assign form (`*ast.TypeSpec.Assign != token.NoPos`).
4. The doc comment matches a compat marker (case-insensitive regex):
   - `backward[- ]?compat`
   - `compat alias`
   - `legacy alias`
   - `prefer X in new callers`
   - `pre-existing lint baseline`

Both spec-level and decl-level doc comment groups are inspected.

Aliases without a marker stay silent — legitimate re-exports must not be over-flagged.

## Why this matters

If the role rule's cheapest remedy is "rename + alias", the rule has produced its own anti-pattern. This companion catches the rule's own collateral. **Self-detecting failure mode is the strongest possible form of rule honesty.**

## Wiring

- `rolealias.go`: new `AnalyzerRoleCompatAlias`.
- Severity `Warning` in `pkg/severity/severity.go`.
- Short + verbose templates in `pkg/messages/struct.go`. Verbose lays out the Goodhart pattern, lists the carve-outs (PNPT/constructor exemptions, marker-less aliases), and gives concrete examples.
- Registry: added to `internal/analyzer/ktn/structure/organization/Analyzers()`.
- `structrolecompatalias_basic` fixture: bad.go has `Sentinel = SentinelService` with backward-compat marker (flagged); good.go has the same shape without a marker (silent).

The structure-level registry test bumps `expectedCount` 21 → 22 and adds `ktnstructrolecompatalias` to the expected name set.

## Out of scope

- Cross-package aliases (`type X = pkg.Y`) — deferred.
- Workspace-wide compat-alias remediation script — separate command if/when needed.

## Test plan

- [x] `make build` green.
- [x] `make test` 97/97.
- [x] `TestStructRoleCompatAlias` — 1 bucket, expected 1.
- [ ] CI verifies the same on the PR branch.

References: issue #263, plan `.claude/plans/issue-263-struct-role-redesign.md`.


kodflow:
**Design rationale on Qodo's "Compat alias requires doc marker" thread**

The marker requirement is intentional, not a gap. The plan locked in this contract before the PR landed:

> Aliases without a marker stay silent — legitimate re-exports must not be over-flagged.

The Goodhart canary is a *self-disclosing* signal: the rule fires when authors admit (in the doc comment) that the alias exists to satisfy the role rule. Flagging every `type X = Y` regardless of doc would over-flag legitimate cross-package re-exports that test packages, generated code, and `internal/` boundary aliases use routinely. That false-positive cost was deemed higher than the benefit of catching marker-less Goodhart fixes.

If a future iteration wants to broaden coverage, it should ship a *separate* opt-in rule (`KTN-STRUCT-ROLE-CROSSPKG-ALIAS` or similar) so the Goodhart canary keeps its narrow self-disclosing semantics.

The other Qodo bug on this PR (Missing base-name check) was a real correctness improvement and was addressed in the latest commit. Thread should auto-resolve once CodeRabbit re-reviews.

PR #270, opened 2026-05-07
fix(stutter): rename 19 stuttering struct types per KTN-STRUCT-ROLE-STUTTER

kodflow:
## Summary

- Drops redundant package-name prefixes/suffixes from 19 exported struct names that stutter at every call site (`user.UserService`, `messages.MessageValue`, `config.RuleConfig`, …) — flagged by the new `KTN-STRUCT-ROLE-STUTTER` analyzer.
- Renames span `pkg/{messages,config,rules,formatter,orchestrator,serve,updater,explore,prompt}` and `internal/analyzer/ktn/tests/{testmeta,classify}`.
- Constructors (`NewXxx`) and ~2700 caller references updated via word-boundary sed; affected `CLAUDE.md` docs refreshed.
- `pkg/config`'s trio keeps the `Spec` suffix (Performance / Rule / Phase alone would have no recognised role and re-trigger `KTN-STRUCT-ROLE`).

## Test plan

- [x] `make build` (Bazel) — green
- [x] `make test` (Bazel) — 97/97 test targets pass
- [x] `mcp__ktn-linter scan phases [1..7]` — 0 issues (was 19)
- [ ] CI pipeline green

- Large API/rename sweep: ~19 exported struct/type names were simplified to remove package-role stutter (examples: RuleInfoSpec→InfoSpec, RuleConfig→RuleSpec, ClassifyResult→Result, MessageValue→Value, PromptOutputResult→OutputResult, OrchestratorService/ServerService→Service, FormatterOptions→Options). ~2,700 caller sites and NewXxx constructors updated; docs (CLAUDE.md, README) and Bazel deps updated to match.
- New/modified KTN rules & message registry changes: KTN-ERROR-DISCARD broadened to treat “error-like” named types and now explicitly applies in test files; README/CLAUDE rule counts updated (KTN rules 200→208, total analyzers 407→415) and a new test rule KTN-TEST-VACUOUS-IDENTITY added. Diagnostic/message wiring moved toward messages.Value and callers updated accordingly.
- Govet/modernize & registry adjustments: wrapper/registry tests and analyzer identity checks were revised—many registry suites now enforce bidirectional expected↔registered bijections and some Doc/Run non-nil assertions were intentionally relaxed where redundant; govet/modernize-style expectations adapted to new analyzer contracts.
- Massive test hardening & coverage delta: hundreds of new white-box unit tests added across analyzers (iota, multinamecount, cargoparams, colocate, iface, goroutine, many more), many existing tests converted to table-driven form. Tests now fail on inspect/type-check errors (no silent ignores), use serialized config/exit-hook guards, and share helpers (mustPipe, drainPipeBuf, mustParseTestFile, isGoModuleCwd), reducing flakes and eliminating numerous false positives.
- CLI/orchestrator and runtime robustness changes: command internals switched to new Service factories (orchestrator.NewService, serve.NewService, updater.NewService, explore.NewService); prompt/explore unified on prompt.OutputResult; formatter/JSON rule payloads use InfoSpec; serve/start helpers gain context-aware early-return and cancellation tests; terminal-restore and exit-hook behavior hardened with new tests. CI pending but local Bazel build and tests are green; ktn-linter scan reports 0 KTN-STRUCT-ROLE issues (was 19).


kodflow:
## Round 3 — pushed `10e9e675`

**15 LEGITIMATE fixes applied** across 13 files. Branch state:
- `./builds/ktn-linter lint ./...` → 0 issues
- Bazel build green

### Highlights
- `testmeta`: refactored `runTestMeta` to a single `insp.Preorder` walk (matches the documented single-pass requirement).
- `serve`: added `blockingServerRunner` so `Test_runServerWithSignalsCtx_cancellation` actually validates cancellation propagation.
- `skill` / `skill_comments`: idempotent registration guards via `cobra.Find` — exercises the registration path without panicking on duplicate `AddCommand`.
- `cliflagsetguard`: added the missing `enabled + missing inspector` test case to cover the `useInsp` branch.
- `errlast`, `nilreturn`, `blankparam`, `unusedarg`, `useclear` (partial): converted tolerant `t.Logf` to `t.Fatalf` after `conf.Check` for fixtures that type-check cleanly.
- `discard_external_test`: refactored `TestErrorDiscard_AppliesInTestFiles` to table-driven `t.Run` pattern.
- `retnaming`: added `wantDiag` tracking — disabled-path test now fails if diagnostics fire while `(nil, nil)` is returned.
- `unusedarg`: extended mutex coverage to two config-mutation paths that were missing the lock.
- `CLAUDE.md`: rule 8 doctrine path now points to the in-repo `internal/analyzer/ktn/tests/CLAUDE.md`.

### Findings rejected with documented justification

**1. `testmeta/result.go` — keep `Result` as struct wrapper (not bare map type)**
The struct wrapper is intentional and documented in `internal/analyzer/ktn/tests/testmeta/CLAUDE.md`:

> The struct wrapper exists so the analyzer's `ResultType` is a named pointer type rather than a bare map alias.

Switching to `type Result map[string]*Value` would break every `result.Metas[name]` consumer across the test/quality/parallel rule packages. Will be revisited only if the documented rationale is overturned.

**2. Four `internal/...` test files importing `pkg/config` (iota, strenum, filename, placement)**
Cherry-picking these 4 files would be inconsistent: the codebase has **489 `internal/ → pkg/config`** imports already (count from `grep -rn "github.com/kodflow/ktn-linter/pkg/config" /workspace/internal/ | wc -l`). The source files under analysis (e.g. `iota.go`, `strenum.go`) themselves import `pkg/config` for the `*config.Config` type. The directional rule in `internal/CLAUDE.md` is aspirational and tracked in the `plan-a-restructure-mecanique-v2` follow-up. Fixing 4 tests while leaving 485+ violations creates churn without architectural benefit.

**3. `t.Skip` instead of silent `return` in `coverage_internal_test.go`**
`KTN-TEST-NOSKIP` (Warning) explicitly forbids `t.Skip()` in this codebase. The `t.Logf("skipped: ...") + return` pattern is the project-sanctioned alternative for env-gated tests. The current sites already log on bypass (`coverage_internal_test.go:1870` carries an explanatory `//:` comment) so coverage is not silent.

**4. `isGoModuleCwd` heuristic in `lint_internal_test.go`**
Deferred. Improving the heuristic to walk the module tree for any reachable `*.go` is a behaviour change that affects both `go test` and Bazel hermetic skipping. Without parallel coverage of both environments, tightening the predicate risks turning a silent skip into a Bazel hard-fail. Will be addressed in a dedicated PR with both runtimes verified.

@coderabbitai review

PR #324, opened 2026-05-15
fix(KTN-GOROUTINE-DOUBLECLOSE): verify builtin close identity and atomic.Bool guard dominance

kodflow:
## Summary

Fixes two correctness bugs in `internal/analyzer/ktn/goroutine/safety/doubleclose.go`:

- **`isCloseBuiltin`**: matched `ident.Name == \"close\"` only, producing a false positive on a local function shadowing the builtin. Now resolves through `pass.TypesInfo.Uses[ident]` and verifies the object is `*types.Builtin` with `Name() == \"close\"`.
- **`hasAtomicBoolGuard`**: accepted `CompareAndSwap`/`Swap` by name with no receiver-type check, anywhere in the function body. Now V1-strict per plan A.4:
  - receiver must resolve to `*sync/atomic.Bool` via TypesInfo
  - only two syntactic forms accepted, with `close(ch)` directly under the `if` body:
    - `if closed.CompareAndSwap(false, true) { close(ch) }`
    - `if !closed.Swap(true) { close(ch) }`
  - other atomic families (Int*, Uint*, Value, Pointer) and general path dominance remain out of scope

Refactored `pkg/serve/watcher.go` and `pkg/serve/bolt_store.go` `Close()` methods to use the strict form so `make lint` stays clean.

## Test plan
- [x] `bazel test //internal/analyzer/ktn/goroutine/safety/...` passes (1/1)
- [x] `go test ./internal/analyzer/ktn/goroutine/safety/...` 345/345 pass
- [x] `go test ./pkg/serve/...` passes (Close() refactor)
- [x] `./builds/ktn-linter lint --only-rule KTN-GOROUTINE-DOUBLECLOSE ./...` returns 0 issues
- [x] `./builds/ktn-linter lint --phases 1,2,3,4,5,6,7 ./internal/analyzer/ktn/goroutine/safety/` returns 0 issues
- [x] testdata covers: shadowed-close, strict CAS form, strict Swap form, atomic.Int32 (must fire), nested under inner FuncLit (must fire)

Closes #308

## Fix for KTN-GOROUTINE-DOUBLECLOSE: stricter builtin close detection and atomic guard validation

- **Rule fix (KTN-GOROUTINE-DOUBLECLOSE)**: Enhanced `isCloseBuiltin()` to resolve identifiers via `TypesInfo` and verify the resolved object is `*types.Builtin` named "close", eliminating false positives when local identifiers shadow the builtin. Tightened `hasAtomicBoolGuard()` to accept only two strict syntactic forms on `*sync/atomic.Bool` where `close(ch)` is directly under the if body: `if closed.CompareAndSwap(false, true) { close(ch) }` and `if !closed.Swap(true) { close(ch) }`. Other atomic families (Int32, Pointer, Value, etc.) and general path-dominance are out of scope for V1.

- **Test coverage expansion**: Updated external test expectation from 2 to 4 issues, enabling detection of previously-missed cases (atomic.Int32 guards and indirectly-nested close calls). Added internal unit tests for builtin resolution with and without TypesInfo fallback, plus fixture-driven tests covering shadowed-close, strict CAS/Swap guard forms, invalid atomic types, and nested FuncLit scenarios.

- **Production code alignment**: Refactored `pkg/serve/watcher.go` and `pkg/serve/bolt_store.go` Close() methods to use the strict `CompareAndSwap(false, true)` guard form, moving shutdown logic into the if body to comply with the analyzer's V1 requirements.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kodflow/ktn-linter/pull/324?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


kodflow:
Re: Qodo 'Missing badShadowedClose fixture' (https://github.com/kodflow/ktn-linter/pull/324#discussion_r3251524565)

**Rebut — non-legit per project convention.**

CLAUDE.md mandates that testdata fixtures expected to produce 0 diagnostics live in `good.go` with the `good*` prefix, and only fixtures with `// want` markers live in `bad.go` with the `bad*` prefix. The shadowed-close fixture asserts the analyzer must NOT fire (TypesInfo correctness check), so by definition it belongs in `good.go` as `goodShadowedClose`. Renaming it to `badShadowedClose` would put a 0-diagnostic function under a `bad*` prefix and break the convention every other rule in the repo follows.

The Qodo 'PR Compliance ID 10' name expectation is the inverted form; the semantic intent (must-not-fire regression for shadowed builtin) is fully covered by the current `goodShadowedClose` in `good.go`. CI green, PR is MERGEABLE.

Refs #308.

PR #345, opened 2026-05-21
fix(release): self-healing asset pipeline, no CSV, dual naming

kodflow:
## Summary

Two coupled fixes restoring `ktn-linter upgrade` and the devcontainer feature install for every existing and future release, **without a hand-maintained mapping file**.

### Commit 1 — `fix(release): auto-build assets on release-please tags + dual naming`

- **release-please.yml** chains into `rebuild-release.yml` as `workflow_call` when `release_created == true`. Future releases get their binaries uploaded automatically — fixes the silent "tag without assets" regression introduced by the renumerology cutover.
- **rebuild-release.yml** drops the hand-maintained `.release/version-map.csv` lookup and auto-detects the build system from the tag's own checkout (`MODULE.bazel` → bazel, `go.mod` → go, neither → placeholders). Commit SHA derived from `git rev-parse HEAD` on the checked-out tag. Adds `workflow_call` trigger alongside the existing `workflow_dispatch`.
- **build-bazel-fallback.sh** / **build-go-fallback.sh** mirror each built binary into `dist/ktn-linter-<goos>-<goarch>` (kebab-case raw, `.exe` on windows) alongside the existing `ktn-linter_<goos>_<goarch>.{tar.gz,zip}`. `pkg/updater/updater.go:getBinaryName()` downloads the raw form directly with no extraction step; the devcontainer feature install.sh keeps using the tarball — both consumers now resolve a 200.

### Commit 2 — `fix(release): kill version-map.csv dependency in validate-releases.sh`

- **validate-releases.sh** rewritten to source from `git tag -l 'v*.*.*'` + `gh release view` instead of iterating `.release/version-map.csv`. Every tag that exists on the remote is now audited, including post-renumerology v1.32+ that were never appended to the CSV — the actual root cause of the silent failure.
- `required_assets` matrix expanded to **11 names** (5 snake tarballs + 5 kebab raw + checksums.txt) so the audit gate matches what consumers actually download.
- `force` flag adapts to state: `force=false` for missing release (fresh create), `force=true` for incomplete release (overwrites the shell). Fixes the 8 consecutive failures on v1.0.x where `force=false` collided with existing releases.
- `INCLUDE_PRERELEASES=true` opt-in for RC tags (off by default — rc.yml owns RC).

### Resulting behaviour

| Path | Trigger | Action |
|---|---|---|
| Happy path | release-please cuts tag → fires the `build-assets` job in the same workflow run | Tag gets all 11 assets automatically |
| Self-healing | `validate-releases.yml` cron `0 * * * *` sweeps every stable tag | Re-dispatches rebuild for any release missing/incomplete |
| Manual | `gh workflow run rebuild-release.yml -f tag=vX -f force=true` | Same as before, still works |

`.release/version-map.csv` is no longer consulted by any code path. Stays on disk for historical reference; deletable in a follow-up subtraction PR.

Full V2 architecture rationale + migration plan: `.claude/contexts/release-pipeline-v2.md` (session-local research artefact, gitignored).

## Test plan

- [ ] CI green on this PR (workflows + scripts don't touch the lint/test code paths)
- [ ] After merge, **re-dispatch** the 3 currently-coquille releases against `--ref main` (the first round of backfills ran against `--ref fix/release-assets-pipeline` but `tooling/` is always checked out from main → got the OLD build scripts; only snake-case tarballs were uploaded):
  ```
  gh workflow run rebuild-release.yml -f tag=v1.32.0 -f force=true
  gh workflow run rebuild-release.yml -f tag=v1.32.1 -f force=true
  gh workflow run rebuild-release.yml -f tag=v1.33.0 -f force=true
  ```
- [ ] After that, both naming conventions present:
  ```
  curl -fI https://github.com/kodflow/ktn-linter/releases/download/v1.33.0/ktn-linter-linux-amd64       # → 200
  curl -fI https://github.com/kodflow/ktn-linter/releases/download/v1.33.0/ktn-linter_linux_amd64.tar.gz # → 200
  ```
- [ ] `./builds/ktn-linter upgrade --check` no longer reports `download failed: status 404`
- [ ] Hourly `validate-releases.yml` cron run logs `ok: vX.Y.Z` for every stable tag (no more `missing release` / `incomplete release` spam)

## Current backfill status (run during PR development)

- v1.32.0, v1.32.1, v1.33.0 all received the **snake-case tarballs** + checksums via 3 manual `gh workflow run` against this branch (workflow yaml = new, build scripts = main's old version)
- Kebab-case raw binaries are NOT yet uploaded — needs a re-dispatch against `--ref main` after this PR merges

[Review Change Stack](https://app.coderabbit.ai/change-stack/kodflow/ktn-linter/pull/345?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


kodflow (commented):
Triage done in commit pushed at 2026-05-21T21:48Z. 6 findings fixed (template injection on all interpolations, grep regex bug, dispatch exit-code propagation, zip-outside-tmp race, permissions justification comment), 1 rejected with reason (SHA-pinning is repo-wide and out of scope here — see PR comment for details). Awaiting CodeRabbit re-review on the new commit.

